PR #5, opened 2024-04-08
HOWTO: Read State from the RPC

### LINKED ISSUE #4: Write a HOWTO for reading contract state
ADR1 abstractly describes how to craft a State, by reading a contract's data via RPC calls and the SDK. A more detailed HOWTO, explaining this process with concrete code would be helpful.

Kukovec:
closes #4

### DIFF
--- a/doc/HOWTO/howto_readState.md
+++ b/doc/HOWTO/howto_readState.md
@@ -159,6 +159,7 @@ https://github.com/stellar/soroban-examples/blob/v20.2.0/increment/src/lib.rs
 [lab]:
 https://laboratory.stellar.org/#xdr-viewer?type=TransactionEnvelope&network=test
 
-[sdk]: https://stellar-sdk.readthedocs.io/en/latest/_modules/stellar_sdk/xdr/ledger_entry_data.html#LedgerEntryData
+[sdk]: https://stellar.github.io/js-stellar-sdk/SorobanDataBuilder.html#.fromXDR
+
 
 
--- a/doc/HOWTO/howto_readState.md
+++ b/doc/HOWTO/howto_readState.md
@@ -1,0 +1,164 @@
+# HOWTO: Read State from the RPC
+
+## Goals 
+In [ADR1](../ADR/ADR_001.md) we described how we wanted to use an intermediate representation of State, to work with within Solarkraft. 
+
+This HOWTO breaks down the various steps required to programmatically extract State in Javascript/Typescript, by using Soroban RPC calls.
+
+## Process
+
+We will be using [this contract][contract] in our example. Its internal state is composed of a single `COUNTER` of type `u32`.
+
+To start with, we will need the deployed contract's ID. To quote from [this guide][guide]:
+
+> When you deploy a contract, first the code is "installed" (i.e., it is uploaded onto the blockchain). This creates a LedgerEntry containing the Wasm byte-code, which is uniquely identified by its hash (that is, the hash of the uploaded code itself). 
+
+Next, we need to identify the value, or set of values, we wish to read. In our case, this is a single field, named `CONTRACT`. 
+
+For each of these values, we first need to determine under which key they are stored in the ledger. This involves computing an XDR representation.
+
+The following snippet shows a function, which can be used to compute this for any contract ID and any field name (using the `stellar-sdk`):
+```js
+const getLedgerKeySymbol = (contractId, symbolText) => {
+  const ledgerKey = xdr.LedgerKey.contractData(
+    new xdr.LedgerKeyContractData({
+      contract: new Address(contractId).toScAddress(),
+      key: xdr.ScVal.scvSymbol(symbolText),
+      durability: xdr.ContractDataDurability.persistent(),
+    }),
+  );
+  return ledgerKey.toXDR("base64");
+};
+```
+
+We need to call `getLedgerKeySymbol(contractID, "COUNTER")`. If we had a contract with multiple fields, we would call this method for each of them:
+```
+k = getLedgerKeySymbol(contractID, "COUNTER")
+k_foo = getLedgerKeySymbol(contractID, "FOO")
+k_bar = getLedgerKeySymbol(contractID, "BAR")
+```
+
+With the key, or set of keys, in hand, we now invoke the RPC, to querry the current values of the fields these keys represent.
+
+In Javascript, querying the RPC is done as follows:
+```js
+let res = await fetch(NET, {
+  method: 'POST',
+  headers: {
+    'Content-Type': 'application/json',
+  },
+  body: JSON.stringify(requestBody),
+})
+let json = await res.json()
+```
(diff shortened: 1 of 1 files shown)


doc/HOWTO/howto_readState.md
```diff
@@ -0,0 +1,164 @@
...
+      }
+      {
+        "key": ...,
+        "xdr": XDR_BAR,
+        "lastModifiedLedgerSeq": ...
+      }
+    ],
+```
+
+The final step is unpacking the obtained XDR, from which we  read the value associated with our key. This [functionality][sdk] is available from the `stellar-sdk` (`LedgerEntryData.from_xdr`), as well as [online][lab] for quick experimentation.
```
konnov, line 119: The link points to python code. Is there code available in JS/TS?

doc/HOWTO/howto_readState.md
```diff
@@ -0,0 +1,164 @@
...
+If the field we're interested in was some other type, say a map, it would look like this:
+```
+...
+val: [scvMap]
+    map: { "foo": 100, "bar": true, ...}
+```
+
+To summarize, these are the steps required to create a State:
+0. Obtain a `contractId`.
+1. Identify a collection of field identifiers (`"FOO", "BAR", "BAZ", ...`) that will make up the State
```
konnov, line 148: Is there a programmatic way to access contract metadata such as the field names and their types?

konnov (dismissed):
Looks nice! I have a few questions, see the comments.

doc/HOWTO/howto_readState.md
Kukovec: I think this would be it: https://stellar.github.io/js-stellar-sdk/SorobanDataBuilder.html#.fromXDR

doc/HOWTO/howto_readState.md
Kukovec: There's a way to get the entire contract bytecode, so I would have to imagine yes. Needs more looking into though.

konnov: shall we expect more HOWTOs on the way?

Kukovec: Ideally one more for tx data, and then a longer one for setting up a testing environment, see #6. 
I imagine this particular question will be resolved in the last part, once I have the local instances set up and can actually fiddle with inputs.

andrey-kuprianov:
A request: for all PRs involving text documents, could you please include the link to the rendered version in the PR description? For this PR it's this [rendered document](https://github.com/freespek/solarkraft/blob/jk/howto1/doc/HOWTO/howto_readState.md). This will simplify the reviewing process.

andrey-kuprianov:
Another request: for all external references, would be nice to have them as links, and not as `code snippets`. E.g. instead of `getLedgerEntries` having [getLedgerEntries](https://developers.stellar.org/network/soroban-rpc/methods/getLedgerEntries) would be tremendously helpful.

doc/HOWTO/howto_readState.md
```diff
@@ -0,0 +1,165 @@
...
+    'Content-Type': 'application/json',
+  },
+  body: JSON.stringify(requestBody),
+})
+let json = await res.json()
+```
+where `NET` is the Soroban network being queried, for example 
+`'https://soroban-testnet.stellar.org'`, and `requestBody` is the specific request we're making of the network.
+
+Since we aim to read a ledger entry, our request will use the `getLedgerEntries` method, and `requestBody` will look like this:
```
thpani, line 57: I still don't think we should limit ourselves to use `getLedgerEntries`, given that it only returns the current state of a contract.

afaiu, the solution for requesting a specific ledger may look very different.

doc/HOWTO/howto_readState.md
```diff
@@ -0,0 +1,165 @@
...
+            contractId: ...
+        key: [scvVec]
+            vec: Array[2]
+                [0]: [scvSymbol]
+                    sym: Counter
+                [1]: [scvAddress]
+                    address: ...
+        durability: {"name":"persistent","value":1}
+        val: [scvU32]
+            u32: VAL
```
thpani, line 136: It looks here like `VAL` is keyed by it's type?

Can you include how we would read `VAL` from this, given either the XDR or whatever the stellar SDK translates it to (I assume JSON)?

thpani (commented):
Things start to make more sense.

As I said in #2, I'm not convinced that retrieving only the latest ledger state is enough.

andrey-kuprianov:
I think the presentation is good! The one thing I miss though is the clear delineation between the example used (the [IncrementContract](https://github.com/stellar/soroban-examples/blob/v20.2.0/increment/src/lib.rs)), and the extensions with more fields / other types. 

I also think the example used is deceptively simple: it's too far from anything used in reality. Not sure if this should be addressed in this HOWTO, or in a separate (advanced) one, but what I would like to see is:

- A HOWTO explaining how to read the state of a non-trivial example; for that I would use the same that is used in the [submission example](https://github.com/freespek/solarkraft/tree/jk/howto1/doc/scf24/example), namely the [Timelock contract](https://github.com/stellar/soroban-examples/blob/v20.0.0/timelock/src/lib.rs)
- Provide a guide of how to read the full state of this specific contract. As it already contains internal structure, no detours like `If we had a contract with multiple fields` needs to be done.
- Provide a fully functional TypeScript code to read and output the state of this specific contract.

Again, not sure this needs to be done within the scope of this HOWTO, or of the subsequent one; only elaborating here what I would ideally like to see:)

Kukovec:
> I think the presentation is good! The one thing I miss though is the clear delineation between the example used (the [IncrementContract](https://github.com/stellar/soroban-examples/blob/v20.2.0/increment/src/lib.rs)), and the extensions with more fields / other types.
> 
> I also think the example used is deceptively simple: it's too far from anything used in reality. Not sure if this should be addressed in this HOWTO, or in a separate (advanced) one, but what I would like to see is:
> 
>     * A HOWTO explaining how to read the state of a non-trivial example; for that I would use the same that is used in the [submission example](https://github.com/freespek/solarkraft/tree/jk/howto1/doc/scf24/example), namely the [Timelock contract](https://github.com/stellar/soroban-examples/blob/v20.0.0/timelock/src/lib.rs)
> 
>     * Provide a guide of how to read the full state of this specific contract. As it already contains internal structure, no detours like `If we had a contract with multiple fields` needs to be done.
> 
>     * Provide a fully functional TypeScript code to read and output the state of this specific contract.
> 
> 
> Again, not sure this needs to be done within the scope of this HOWTO, or of the subsequent one; only elaborating here what I would ideally like to see:)

This is actually part of the last bullet point in #6, and something I'm working on, but it requires me to set up a local network and rpc so I can do experiments, and the installations aren't cooperating with me.

thpani:
We ended up needing [Horizon](https://developers.stellar.org/network/horizon) instead of Soroban RPC

PR #10, opened 2024-04-12
HOWTO: Set Up a Testnet and Deploy Custom Contracts

### LINKED ISSUE #9: Write a HOWTO for the configuration and deployment process

Kukovec:
closes #9 

This HOWTO describes the steps required to:
- Create a local Stellar network
- Write and compile a custom contract
- Deploy said contract to a (test)net and invoke it

[Rendered doc](https://github.com/freespek/solarkraft/blob/jk/howto2/doc/HOWTO/howto_setupAndDeployCustom.md)

### DIFF
--- a/doc/HOWTO/howto_setupAndDeployCustom.md
+++ b/doc/HOWTO/howto_setupAndDeployCustom.md
@@ -1,4 +1,4 @@
-# HOWTO: Set Up a Testnet and Deploy Custom Contracts
+# HOWTO: Set Up a Network and Deploy Custom Contracts
 
 ## Goals 
 This HOWTO details the steps required to set up a local Stellar network, deploy a custom Soroban contract, and invoke the contract via CLI. It complements the [official guide][guide].
@@ -11,7 +11,7 @@ Prerequisites:
 - you probably want to add `~/.cargo/bin` to your PATH
 
 ### Local Stellar Network
-Instead of building from source, Stellar provides a quickstart Docker image. You can call the below command from anywhere, and Docker will fetch the image.
+Instead of building from source, Stellar provides a quickstart Docker image. You can call the below command from anywhere, and Docker will fetch the image and run it in the background.
 ```sh
 docker run --rm -d \
     -p 8000:8000 \
@@ -20,9 +20,9 @@ docker run --rm -d \
     --testnet \
     --enable-soroban-rpc
 ```
-Note here that the [official guide][guide] would have you run this in interactive mode (`run ... -it ...`), but for just the network having it run in the background (`-d`) is less noisy and doesn't hijack the terminal.
+Note here that the [official guide][guide] would have you run the container in interactive mode (`run ... -it ...`), but for just the network having it run in the background (`-d`) is less noisy and doesn't hijack the terminal.
 
-The above snippet deploys a testnet (`--testnet`), if you want a standalone network, replace that flag with `--standalone`.
+The above snippet deploys a new node, which is connected to the global testnet (`--testnet`), if you want a standalone network, replace that flag with `--local` (in outdated documentation you may find `--standalone`, which is a synonym).
 
 ### Soroban CLI
 In order to interact with your docker image, you will need the [Soroban CLI][cli]. With `cargo`, all you need is:
@@ -75,12 +75,17 @@ soroban-sdk = { workspace = true }
 soroban-sdk = { version = "20.3.2", features = ["testutils"] }
 ```
 
-Write your contract code in `lib.rs`, and tests in `test.rs`.
+Write your contract code in `lib.rs`, and tests in `test.rs`. You can start by using one of our [example contracts](../../ContractExamples/contracts
+)
 Make sure all your tests work before deploying (with print outputs):
 ```sh
 cargo test -- --nocapture
 ```
 Once they do, running
+```sh
+rustup target add wasm32-unknown-unknown
+```
+then
 ```sh
 soroban contract build
 ```
--- a/doc/HOWTO/howto_setupAndDeployCustom.md
+++ b/doc/HOWTO/howto_setupAndDeployCustom.md
@@ -1,0 +1,144 @@
+# HOWTO: Set Up a Testnet and Deploy Custom Contracts
+
+## Goals 
+This HOWTO details the steps required to set up a local Stellar network, deploy a custom Soroban contract, and invoke the contract via CLI. It complements the [official guide][guide].
+
+## Process
+
+Prerequisites:
+- Make `cargo`, `rustup`, and `go` available. You will need a`rustc` version that is at least 1.74.
+- Make sure you have enough allocated space. If you're using the NixOS VM, you might need to add `services.logind.extraConfig = "RuntimeDirectorySize=2G";` to your configuration, rebuild and reboot the VM.
+- you probably want to add `~/.cargo/bin` to your PATH
+
+### Local Stellar Network
+Instead of building from source, Stellar provides a quickstart Docker image. You can call the below command from anywhere, and Docker will fetch the image.
+```sh
+docker run --rm -d \
+    -p 8000:8000 \
+    --name stellar \
+    stellar/quickstart:testing \
+    --testnet \
+    --enable-soroban-rpc
+```
+Note here that the [official guide][guide] would have you run this in interactive mode (`run ... -it ...`), but for just the network having it run in the background (`-d`) is less noisy and doesn't hijack the terminal.
+
+The above snippet deploys a testnet (`--testnet`), if you want a standalone network, replace that flag with `--standalone`.
+
+### Soroban CLI
+In order to interact with your docker image, you will need the [Soroban CLI][cli]. With `cargo`, all you need is:
+```sh
+cargo install --locked soroban-cli --features opt
+```
+
+note that this gets placed in `~/.cargo/bin`, which is why you want that as part of your `PATH`.
+
+### Custom Soroban Contracts
+
+<!--- TODO: Make a megacontract and link it here-->
+
+You can use the Soroban CLI to set up a blank project, and add your own contract code to it (you may also optionally include any of their example contracts).
+
+To set up a new _project_ (not contract!), type:
+```sh
+soroban contract init <DIR>
+```
+This will populate `<DIR>` (e.g. `.`) with some basic files, most notably a `Cargo.toml` config. Your custom contract should be added in `<DIR>/contracts/<CONTRACT>`.
+For example, for a contract `Foo`, you'd create:
+```
+<DIR>
+├── Cargo.toml
+└── contracts
+    └── Foo
+        ├── Cargo.toml
+        └── src
+            ├── lib.rs
+            └── test.rs
+```
+
+Note that `Cargo.toml` in `Foo` is _not_ the same `Cargo.toml` in `<DIR>`! The latter gets created automagically with `soroban contract init`, the former you have to add yourself. Here's an example of `Cargo.toml` in the contract folder:
+```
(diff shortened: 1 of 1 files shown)


Kukovec:
In case someone wants a quick contract to play around with, here's a `lib.rs` from `Foo`. It has a map of vectors in addition to an integer counter.
```rust
#![no_std]
use soroban_sdk::{contract, contractimpl, log, symbol_short, Env, Symbol, Vec, Map, Address};

const VAL: Symbol = symbol_short!("VAL");
const ADDRS: Symbol = symbol_short!("ADDRS");

#[contract]
pub struct FooContract;

type AddrMap = Map<u32, Vec<Address>>;

#[contractimpl]
impl FooContract {
    pub fn bar(env: Env, addr: Address) -> u32 {
        // Get the current count.
        let mut val: u32 = env.storage().instance().get(&VAL).unwrap_or(1); // If no value set, assume 0.
        log!(&env, "val: {}", val);

        // If val < 50, double, else set to 1
        val = if val < 5 { 2 * val } else { 1 };

        // Save val.
        env.storage().instance().set(&VAL, &val);

        // Save the caller, for the current counter value
        let mut addrs: AddrMap = env.storage().instance().get(&ADDRS).unwrap_or(AddrMap::new(&env));

        let mut new_vec = addrs.get(val).unwrap_or(Vec::new(&env));
        new_vec.push_back(addr);

        addrs.set(val, new_vec);

        // Save addrs.
        env.storage().instance().set(&ADDRS, &addrs);

        // The contract instance will be bumped to have a lifetime of at least 100 ledgers if the current expiration lifetime at most 50.
        // If the lifetime is already more than 100 ledgers, this is a no-op. Otherwise,
        // the lifetime is extended to 100 ledgers. This lifetime bump includes the contract
        // instance itself and all entries in storage().instance(), i.e, COUNTER.
        env.storage().instance().extend_ttl(50, 100);

        // Return val to the caller.
        val
    }

    pub fn baz(env: Env) -> AddrMap {
        return env.storage().instance().get(&ADDRS).unwrap_or(AddrMap::new(&env));
    }
}

mod test;
```

Kukovec:
As we've discussed, we should create a megacontract in the future, and when we do I'll link it here as an example.

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Process
+
+Prerequisites:
+- Make `cargo`, `rustup`, and `go` available. You will need a`rustc` version that is at least 1.74.
+- Make sure you have enough allocated space. If you're using the NixOS VM, you might need to add `services.logind.extraConfig = "RuntimeDirectorySize=2G";` to your configuration, rebuild and reboot the VM.
+- you probably want to add `~/.cargo/bin` to your PATH
+
+### Local Stellar Network
+Instead of building from source, Stellar provides a quickstart Docker image. You can call the below command from anywhere, and Docker will fetch the image.
```
konnov, line 14: Since you mention Nixos above, we should add a link to setting up docker on Nixos: https://nixos.wiki/wiki/Docker

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+doctest = false
+
+[dependencies]
+soroban-sdk = { workspace = true }
+
+[dev-dependencies]
+soroban-sdk = { version = "20.3.2", features = ["testutils"] }
+```
+
+Write your contract code in `lib.rs`, and tests in `test.rs`.
```
konnov, line 78: I have made till this point. Could you add an example contract in the repo, so all of us could the same one to make sure that we have our setup working.

konnov (commented):
I have made it through the first half of the doc! It's nice and detailed. I left a few comments. Stopped at a blocker that requires me to find a contract somewhere. Please add one.

Kukovec:
> I have made it through the first half of the doc! It's nice and detailed. I left a few comments. Stopped at a blocker that requires me to find a contract somewhere. Please add one.

https://github.com/freespek/solarkraft/pull/10#issuecomment-2051495611

doc/HOWTO/howto_setupAndDeployCustom.md
Kukovec: Might as well take this chance to set up the megacontract repo going forward.

konnov:
> > I have made it through the first half of the doc! It's nice and detailed. I left a few comments. Stopped at a blocker that requires me to find a contract somewhere. Please add one.
> 
> [#10 (comment)](https://github.com/freespek/solarkraft/pull/10#issuecomment-2051495611)

Yeah, that's where I got stuck. Why would you just commit the whole contract layout, e.g., under `examples/contracts`

doc/HOWTO/howto_setupAndDeployCustom.md
Kukovec: https://github.com/freespek/solarkraft/pull/12

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+docker run --rm -d \
+    -p 8000:8000 \
+    --name stellar \
+    stellar/quickstart:testing \
+    --testnet \
+    --enable-soroban-rpc
+```
+Note here that the [official guide][guide] would have you run this in interactive mode (`run ... -it ...`), but for just the network having it run in the background (`-d`) is less noisy and doesn't hijack the terminal.
+
+The above snippet deploys a testnet (`--testnet`), if you want a standalone network, replace that flag with `--standalone`.
```
thpani, line 25: It looks like `--testnet` does not deploy a fresh net, but will join the official testnet?
https://github.com/stellar/quickstart/blob/master/README.md

Also, did you mean `--local` instead of `--standalone`?

Kukovec: No, `--standalone` is what the [admin guide](https://developers.stellar.org/network/soroban-rpc/admin-guide) lists (and `--standalone` works, I've tested it).
RE testnet, you're correct, the text is sloppy, it should accurately read "testnet node".

thpani: Okay, `--standalone` and `--local` are synonyms, but `--standalone` is deprecated:

https://github.com/stellar/quickstart/blob/696ccb26643a8d7052cfbb56f767356007950800/start#L126-L127

Kukovec: Ah lovely, their docs are out of date.

thpani: Your chance for a "stellar" PR!

doc/HOWTO/howto_setupAndDeployCustom.md
thpani: There is [`soroban contract init --with-example`](https://github.com/stellar/soroban-cli/blob/main/docs/soroban-cli-full-docs.md#options-25), you could just use one of those.

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Write your contract code in `lib.rs`, and tests in `test.rs`.
+Make sure all your tests work before deploying (with print outputs):
+```sh
+cargo test -- --nocapture
+```
+Once they do, running
+```sh
+soroban contract build
```
thpani, line 85: I had to run the following before this:

```sh
rustup target add wasm32-unknown-unknown
```

thpani: Ah, it's actually in the setup section of the official guide, so probably necessary everywhere: https://developers.stellar.org/docs/smart-contracts/getting-started/setup#install-the-target

thpani, line 85: Weird... I needed to pass `--out-dir someDir` here, otherwise it wouldn't create the `.wasm` file.

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+
+Next, we create agents within this network. To do this, run
+```sh
+soroban keys generate <AGENT> --network <NETWORK>
+```
+followed by
+```sh
+soroban config identity fund <AGENT> --network <NETWORK>
+```
+The latter will show a warning, saying that `<AGENT>` already exists, but if you call this command before the previous one it fails to execute, and if you don't call it at all the next command fails, so just ignore the warning.
```
thpani, line 110: afaict `soroban keys generate` fails with the funding error if it cannot make a connection to the node. Otherwise, the `config identity fund` is not needed.

thpani (commented):
Yeah, got my first contract deployed! Thanks Jure!

I left I few comments that I needed to get everything going, I think we should wait for someone else to follow along and see if those are quirks of my local setup or needed in general.

doc/HOWTO/howto_setupAndDeployCustom.md
Kukovec: Hmm, I must have done it but not noted it down and forgotten about it. Thanks for pointing this out

Kukovec: Did you use the megacontract folder or was this a from-scratch project? And if you did use it, did you run `soroban contract init` anyway?

thpani: I did a clean start, and now it works. Maybe I ran the `rustup target add wasm32-unknown-unknown` at the wrong point in time 🤷

andrey-kuprianov: Also ran into the same issue, on Linux. The same command solved the issue.

```suggestion
rustup target add wasm32-unknown-unknown
soroban contract build
```

doc/HOWTO/howto_setupAndDeployCustom.md
```diff
@@ -0,0 +1,144 @@
...
+
+Having defined an agent, you can have them deploy the contract with:
+```sh
+soroban contract deploy --wasm <CONTRACT>.wasm --source <AGENT> --network <NETWORK>
```
andrey-kuprianov, line 114: I had to provide also the passphrase, otherwise the command fails:
```suggestion
soroban contract deploy --wasm <CONTRACT>.wasm --source <AGENT> --network <NETWORK> --network-passphrase <PASSPHRASE>
```

andrey-kuprianov: In fact for my setup, this step was failing with 

```error: provided network passphrase "soroban testnet" does not match the server: "Test SDF Network ; September 2015"```

Only when the previous step was executed exactly with this passphrase, i.e.

```soroban config network add testnet --rpc-url "http://localhost:8000/soroban/rpc" --network-passphrase "Test SDF Network ; September 2015"```

and then all arguments repeated in the deploy call like that

```soroban contract deploy --wasm target/wasm32-unknown-unknown/release/hello_world.wasm --source alice --network testnet --network-passphrase "Test SDF Network ; September 2015" --rpc-url "http://localhost:8000/soroban/rpc```

it did succeed...

Kukovec: I think this might be related to https://github.com/freespek/solarkraft/pull/10/files#r1565784375, since the node is joining an existing testnet. It should work as is written with `--local`.

thpani:
We got this working for the dev team quite some time ago, and the Soroban "getting started" docs are probably a more up-to-date resource: https://developers.stellar.org/docs/smart-contracts/getting-started

PR #12, opened 2024-04-12
Contract examples

### LINKED ISSUE #11: Create a collection of contracts for testing as well as a megacontract

Kukovec:
Adds an example contract collection, with a project config.

To discuss: Currently (as seen in `.gitignore`), GH will not track soroban configs (`.soroban`). Whenever you add an agent or network, for example `alice`, there is a file that gets generated in `.soroban` with the relevant data for that agent or network. If we wanted to, we could maintain a shared collection of agents (and networks) in the repo as well.

closes #11


thpani:
Are the `.toml` files autogenerated?

Kukovec:
> Are the `.toml` files autogenerated?

The top-level one yes, the one inside the subdirectory no.

konnov (approved):
Looks good to me

konnov:
> To discuss: Currently (as seen in `.gitignore`), GH will not track soroban configs (`.soroban`). Whenever you add an agent or network, for example `alice`, there is a file that gets generated in `.soroban` with the relevant data for that agent or network. If we wanted to, we could maintain a shared collection of agents (and networks) in the repo as well.

This is a good question. I think we should ignore all files with potentially sensitive data by default. Also, we need a testing environment, where we are only using well-known private-public keys.

Btw, an agent is a bit unusual terminology for smart contracts. Is it Soroban terminology?

Kukovec:
> Btw, an agent is a bit unusual terminology for smart contracts. Is it Soroban terminology?

Not really, it's all just addresses. I just personally find it helpful to distinguish between addresses that just exist, and addresses under your control that you can use as message senders/function invokers

PR #21, opened 2024-04-15
Add a dev container for Soroban development

thpani:
Add a [VSCode dev container](https://code.visualstudio.com/docs/devcontainers/containers) for Soroban development.

This simply takes the Rust dev container from Microsoft and installs `soroban-cli` and the appropriate wasm target.

If the [quickstart Docker image](https://github.com/stellar/quickstart) is run outside the container, its RPC endpoint is available as `http://host.docker.internal:8000/soroban/rpc`.


konnov (approved):
Nice!

Issue #23, opened 2024-04-16
Create a basic representation of smart contract builtin types


Kukovec:
See [this](https://developers.stellar.org/docs/learn/smart-contract-internals/types/built-in-types) list of types

PR #26, opened 2024-04-16
Basic values and tests

Kukovec:
Partially addresses #23 (missing compound values, e.g. Map/Vec)

### DIFF
--- a/solarkraft/src/state/value.ts
+++ b/solarkraft/src/state/value.ts
@@ -53,12 +53,16 @@ export function isValid(v: Value): boolean {
  * 
  * Example: 2u32 would be represented as { val: 2, type: "u32" }, whereas 2i32 would be { val: 2, type: "i32" }
 */
+
+export type ValidIntT = "u32" | "i32" | "u64" | "i64" | "u128" | "i128"
+
 export type IntValue = {
     val: bigint
-} & (IntValue_u32 | IntValue_i32 | IntValue_u64 | IntValue_i64 | IntValue_u128 | IntValue_i128)
+    type: ValidIntT
+}
 
 
-function mkInt(type: string, val: bigint) {
+function mkInt(type: ValidIntT, val: bigint) {
     const obj = { type: type, val: val } as Value
     if (!isValid(obj)) {
         throw new RangeError(`${val} lies outside the ${type} range.`)
@@ -66,55 +70,24 @@ function mkInt(type: string, val: bigint) {
     return obj
 }
 
-
-// u32
-export type IntValue_u32 = {
-    type: "u32"
-}
-
 export function u32(v: bigint): Value {
     return mkInt("u32", v)
-}
-
-// i32
-export type IntValue_i32 = {
-    type: "i32"
 }
 
 export function i32(v: bigint): Value {
     return mkInt("i32", v)
 }
 
-// u64
-export type IntValue_u64 = {
-    type: "u64"
-}
-
 export function u64(v: bigint): Value {
     return mkInt("u64", v)
-}
-
-// i64
-export type IntValue_i64 = {
-    type: "i64"
 }
 
 export function i64(v: bigint): Value {
     return mkInt("i64", v)
 }
 
-// u128
-export type IntValue_u128 = {
-    type: "u128"
-}
-
 export function u128(v: bigint): Value {
     return mkInt("u128", v)
-}
-
-// i128
-export type IntValue_i128 = {
-    type: "i128"
 }
 
 export function i128(v: bigint): Value {
@@ -166,7 +139,7 @@ export function addr(s: string): Value {
 // The `len` field is present iff the length is fixed (i.e. for BytesN)
 // TODO: tests
 export type ArrValue = {
-    val: IntValue_u32[]
+    val: { type: "u32", val: bigint }[]
     type: "arr"
     len?: number
     // if (typeof (l) !== 'undefined' && v.length !== l) {
--- a/solarkraft/src/state/value.ts
+++ b/solarkraft/src/state/value.ts
@@ -1,0 +1,146 @@
+interface Value {}
+
+/**  
+ * Any of the follwing:
+ *  - Unsigned 32-bit Integer (u32)
+ *  - Signed 32-bit Integer (i32)
+ *  - Unsigned 64-bit Integer (u64)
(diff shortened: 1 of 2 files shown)


konnov (commented):
Can we avoid using abstract classes and classes here? There is no need for them. An idiomatic way in TypeScript is to rather use discriminated unions: `{ kind: 'int32', val: bigint } | { kind: 'bool', val: boolean }`.

The reason is that we would run into multiple issues with classes such as `instanceof`, equality, hashing, etc.

konnov:
Also, it's quite easy to construct values with discriminated unions like `{ kind: 'bool', val: true }`, instead of instantiating classes.

Kukovec:
> Can we avoid using abstract classes and classes here? There is no need for them. An idiomatic way in TypeScript is to rather use discriminated unions: `{ kind: 'int32', val: bigint } | { kind: 'bool', val: boolean }`.
> 
> The reason is that we would run into multiple issues with classes such as `instanceof`, equality, hashing, etc.

And how do you enforce field validity in those discriminated unions? The class constructors make sure u32 values are actually u32, I don't see a way of doing that with your suggestion.

konnov:
> > Can we avoid using abstract classes and classes here? There is no need for them. An idiomatic way in TypeScript is to rather use discriminated unions: `{ kind: 'int32', val: bigint } | { kind: 'bool', val: boolean }`.
> > The reason is that we would run into multiple issues with classes such as `instanceof`, equality, hashing, etc.
> 
> And how do you enforce field validity in those discriminated unions? The class constructors make sure u32 values are actually u32, I don't see a way of doing that with your suggestion.

I am actually not sure about why you want to have field validation there. You definitely do not want validation in many cases, e.g., when copying values. This is typically done somewhere in the parser/reader. If you need validation, simply introduce a validation function that does a switch over different kinds. The intermediate representation is just data.

Kukovec:
> > > Can we avoid using abstract classes and classes here? There is no need for them. An idiomatic way in TypeScript is to rather use discriminated unions: `{ kind: 'int32', val: bigint } | { kind: 'bool', val: boolean }`.
> > > The reason is that we would run into multiple issues with classes such as `instanceof`, equality, hashing, etc.
> > 
> > 
> > And how do you enforce field validity in those discriminated unions? The class constructors make sure u32 values are actually u32, I don't see a way of doing that with your suggestion.
> 
> I am actually not sure about why you want to have field validation there. You definitely do not want validation in many cases, e.g., when copying values. This is typically done somewhere in the parser/reader. If you need validation, simply introduce a validation function that does a switch over different kinds. The intermediate representation is just data.

Can you give me an example of such copying, where you would explicitly not want the validity check, because I genuinely don't understand what use case you have in mind.

konnov:
> [...]
> 
> Can you give me an example of such copying, where you would explicitly not want the validity check, because I genuinely don't understand what use case you have in mind.

I don't want to get into a discussion between classes and discriminated unions in TypeScript. What I can say is that there are very few cases, where I have seen value in using classes (I mean, TypeScript classes) in TypeScript. Most of the time, plain objects and discriminated unions do the job just fine. Here are just a few things that make TS classes hard to use: useless equality, useless `toString()` by default, no good support for `instanceof` (you still have to introduce the kind field).

In your PR, we need something similar to sealed classes in Scala. Discriminated unions provide you with exactly that. You don't want to inherit from values, nor you need other features of classes.

solarkraft/src/state/value.ts
```diff
@@ -11,133 +53,125 @@ interface Value { }
  * 
  * Example: 2u32 would be represented as { val: 2, type: "u32" }, whereas 2i32 would be { val: 2, type: "i32" }
 */
-abstract class IntValue implements Value {
+export type IntValue = {
     val: bigint
-    type: string
+} & (IntValue_u32 | IntValue_i32 | IntValue_u64 | IntValue_i64 | IntValue_u128 | IntValue_i128)
```
konnov, line 58: Why would not you simply enumerate all cases right here? You still have to define types for `IntValue_u32` below.

For instance, this is just immediate and does not require any scrolling:

```ts
export type IntValue =
    { val: bigint, type: "u32" }
  | { val: bigint, type: "i32" }
  | { val: bigint, type: "u64" }
  ...
```

Alternatively, we could do:

```ts
export type IntValue =
  { type: "uint", bits: 32 | 64 | 128, val: bigint }
  { type: "int", bits: 32 | 64 | 128, val: bigint }
```

solarkraft/src/state/value.ts
```diff
@@ -11,133 +53,125 @@ interface Value { }
...
-    constructor(v: bigint) {
-        this.val = v
-        if (!this.isValid()) {
-            throw new RangeError(`${v} lies outside the ${this.type} range.`)
-        }
-    }
 
-    abstract isValid(): boolean
+function mkInt(type: string, val: bigint) {
+    const obj = { type: type, val: val } as Value
```
konnov, line 62: this does not seem type-safe to me.

```suggestion
function mkInt(type: "u32" | "u64" | ... | "u128", val: bigint) {
    const obj = { type: type, val: val }
```

solarkraft/src/state/value.ts
```diff
@@ -11,133 +53,125 @@ interface Value { }
...
 }
 
+
 // u32
-export class IntValue_u32 extends IntValue {
-    val: bigint
-    type = "u32"
+export type IntValue_u32 = {
+    type: "u32"
+}
```
konnov, line 73: what is the value of these definitions?

solarkraft/src/state/value.ts
```diff
@@ -11,133 +53,125 @@ interface Value { }
...
     }
+    return obj
 }
 
 // Byte arrays (Bytes, BytesN)
 // The `len` field is present iff the length is fixed (i.e. for BytesN)
 // TODO: tests
-export class ArrValue implements Value {
+export type ArrValue = {
     val: IntValue_u32[]
```
konnov, line 169: if we already know that we have an unsigned `u32` here, why don't we just use `number`? Most likely, you would have to use the array length in loops, which would make it much easier if we are using `number`.

konnov (commented):
I like it really much better! Left a few stylistic comments.

konnov (approved):
Looks good! Let's merge it. I think you would still run into issues when looping over bigints, but we can deal with that later.

Issue #25, opened 2024-04-16
add mock CLI commands


konnov:
as a prototype for our CLI

PR #28, opened 2024-04-16
add a simple CLI mock

konnov:
Towards #25. This PR adds one CLI command `txs`. The main purpose is just to get the argument parser work. I still have to think through the CLI commands.

### DIFF
--- a/solarkraft/eslint.config.js
+++ b/solarkraft/eslint.config.js
@@ -10,4 +10,9 @@ import tseslint from 'typescript-eslint';
 export default tseslint.config(
   eslint.configs.recommended,
   ...tseslint.configs.recommended,
+  {
+    "rules": {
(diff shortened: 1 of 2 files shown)


solarkraft/eslint.config.js
```diff
@@ -10,4 +10,9 @@ import tseslint from 'typescript-eslint';
 export default tseslint.config(
   eslint.configs.recommended,
   ...tseslint.configs.recommended,
-);
\ No newline at end of file
+  {
+    "rules": {
+      "@typescript-eslint/no-explicit-any": "off"
```
konnov, line 15: `yargs` requires `any` and I don't know how to work around this. In general, not allowing `any` at all seems to be very restrictive to me.

thpani (approved):

PR #29, opened 2024-04-18
Add mock CLI commands

### LINKED ISSUE #25: add mock CLI commands

konnov:
Closes #25. I have thought a bit about the CLI commands that we would need. Here is the first mock that contains `fetch`, `list`, and `verify`. To check this PR, simply pull it, compile and play with `solarkraft`.

### DIFF
--- a/solarkraft/src/fetch.ts
+++ b/solarkraft/src/fetch.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
+ */
+/**
+ * Fetch transactions from the ledger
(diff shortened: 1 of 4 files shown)


thpani (commented):
What's the rationale for fetching first, then verifying?

If it's not prohibitively expensive, wouldn't you just fetch on-the-fly during verification?

konnov:
> What's the rationale for fetching first, then verifying?
> 
> If it's not prohibitively expensive, wouldn't you just fetch on-the-fly during verification?

We don't know yet. It may be expensive in some case. For now, it is just to separate our work more efficiently :)

thpani (approved):

PR #31, opened 2024-04-18
Split dev container

thpani:
Building the dev container was rather slow (because `cargo install soroban-cli` is slow).

So I'm splitting it into two separate container definitions:
- TypeScript + Java + Apalache in `./solarkraft/`
- Rust + Soroban in `./ContractExamples/`

I'm also getting rid of `Dockerfile`; there's devcontainer features for both `cargo install` and adding the `wasm32-unknown-unknown` target.

This way, you can quickly bring up container for Solarkraft from scratch, while we have a second one for Soroban-related work.

For the Solarkraft container, Apalache is installed in `/opt/apalache/`.


solarkraft/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,24 @@
+// A dev container for working with Soroban smart contracts.
```
konnov, line 1: The descriptions of the both containers is identical. What's the difference?

konnov (approved):
Nice! Could you also add some instructions to `CONTRIBUTING.md`? I would like to use these containers but not sure about when and how :)

solarkraft/.devcontainer/devcontainer.json
thpani: Aaah, I thought I did changed it 😄 
[e171c2a](https://github.com/freespek/solarkraft/pull/31/commits/e171c2a477c78d06a0b0d3c87519767871d6f56a)

PR #33, opened 2024-04-19
Pin devcontainer platform to linux/amd64

thpani:
There are no Z3 binaries for linux/aarch64 for now: https://github.com/Z3Prover/z3/issues/7201

Pin the devcontainer platform to `linux/amd64` to work around it.


konnov (approved):
LGTM!

PR #36, opened 2024-04-22
ITF serialization of `Value`s

### LINKED ISSUE #35: Write ITF serialization

Kukovec:
closes #35 

Introduces the following:
  - the `State` type
  - A rudimentary type system, notationally based on that of Apalache, with type literals from Soroban (`u32`, `symb`, etc.)
  - An ITF representation (of a single state) as a length-1 trace
  - modifies validity checks on collections, to require homogeneous contents, unless explicitly allowed otherwise.


solarkraft/src/state/value.ts
```diff
@@ -16,7 +16,88 @@ export type Value = {
     | MapValue
 )
 
-export function isValid(v: Value): boolean {
+const EMPTY_COLLECTION_TYPE: string = 'T'
+const UNKNOWN_TYPE: string = 'UNKNOWN'
+
+// helper function
+function computeSingleTypeRep(ts: string[]): string[] {
```
thpani, line 23: I'd have to guess what this does. Can we add a more helpful comment?

solarkraft/src/state/value.ts
```diff
@@ -16,7 +16,88 @@ export type Value = {
...
+        case 'i32':
+        case 'u64':
+        case 'i64':
+        case 'u128':
+        case 'i128':
+        case 'symb':
+        case 'addr':
+            return v.type
+        case 'arr':
+            return 'Bytes'
```
thpani, line 46: How is 'arr' different from the other basic types?

thpani (commented):
A first set of comments.

The ITF-serialization of state looks good.

I've not had time to look at all the code for supporting nonhomogenous collections.

solarkraft/src/io/itf.ts
```diff
@@ -0,0 +1,92 @@
...
+        case 'i64':
+        case 'u128':
+        case 'i128':
+            return numWrap(v.val)
+        case 'symb':
+        case 'addr':
+            return v.val
+        case 'arr':
+            // array of 0/1 s, must be wrapped
+            return v.val.map(numWrap)
```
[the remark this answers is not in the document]
Kukovec: Yes, I did it for consistency, but for `Bytes` specifically, we could deviate from ITF and just use the JS numbers, since it's only 0s and 1s, b/c I do agree that it's a bit clunky.

solarkraft/src/io/itf.ts
```diff
@@ -0,0 +1,92 @@
...
+        case 'vec':
+            return v.val.map(valToITF)
+        case 'map':
+            return {
+                [MAP_FIELD]: Array.from(v.val.entries()).map(
```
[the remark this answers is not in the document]
Kukovec: I tried it, and I don't think it does: "Exception during run: test/io/itf.test.ts(13,86): error TS2339: Property 'map' does not exist on type 'IterableIterator<[string, string]>'."
And even if it did, you want an array there anyway, and I would assume, if it were implemented, that mapping over an iterator would give an iterator.

solarkraft/src/io/itf.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+export function stateToITF(s: State): object {
+    const formatter = new Intl.DateTimeFormat('en-GB', {
+        dateStyle: 'full',
+        timeStyle: 'long',
+    })
+
+    const desc = `Created by Solarkraft on ${formatter.format(new Date().getTime())}`
```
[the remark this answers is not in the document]
Kukovec: I tried this, and the output is:
```
Day NN Month Year at HH:MM:SS CEST
```
in the current version, vs
```
DayAbbreviated Month NN Tear HH:MM:SS GMT+XX00 (Central European Summer Time)
```
in your proposal.

I would still recommend going with an explicit date formatter, for consistency and predicatability, but I don't feel strongly about this. 
For reference, here's the date from an Apalache ITF I had lying around:
```
Mon Apr 22 12:30:12 CEST 2024
```

solarkraft/src/state/value.ts
Kukovec: I'll add a comment about this inline, but I wanted to convey that `Bytes`, while it is a collection, is basically atomic as a type, since it's always an array of exactly 0s or 1s. I felt like using the word `array`/`arr` might give the impression that this is an array in the traditional sense, i.e. a fixed-sized ordered collection of _arbitrary_ values, which is very much not supported (but there is `Vec`).

solarkraft/test/io/itf.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        const s = itf['states'][0]
+
+        const expectedBalances = [
+            [
+                aliceAddr,
+                { '#bigint': ['1'].concat(new Array(20).fill('0')).join('') },
+            ],
+            [
+                bobAddr,
+                { '#bigint': ['1'].concat(new Array(10).fill('0')).join('') },
```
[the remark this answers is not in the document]
Kukovec: Weird, I thought I tried that and it didn't have the method, I must have been doing something else. I'll fix this.

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
konnov: If it's an array of zeroes and ones, why would not we just wrap it as a string, e.g., by introducing a new kinds of a field `{ "#bytes": "101" }`, or by just using an array of numbers, as you are suggesting. Actually, ITF does not prohibit JS numbers from being used for the small ranges.

solarkraft/src/io/itf.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+export function stateToITF(state: State): object {
+    const formatter = new Intl.DateTimeFormat('en-GB', {
+        dateStyle: 'full',
+        timeStyle: 'long',
+    })
+
+    // const desc = `Created by Solarkraft on ${formatter.format(new Date().getTime())}`
```
konnov, line 63:
```suggestion
```

solarkraft/src/state/value.ts
```diff
@@ -16,7 +16,98 @@ export type Value = {
...
+    else {
+        const t0 = ts[0]
+        const hom = !ts.some((t) => t !== t0)
+        if (hom) return [t0]
+        else return ts
+    }
+}
+
+// Returns the full type annotation of the provided (possibly complex) Value `v`.
+// It uses Apalache-style type notation, with Soroban type primitives.
```
konnov, line 40: what is Apalache-style type notation? What we see below looks like Soroban to me.

konnov (commented):
Left a few comments

solarkraft/src/state/value.ts
Kukovec: Well, our "maps" are `A -> B`

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
Kukovec: IIRC we had a discussion about standardizing this in Apalache where _every_ value was `#bigint`. But yeah, we don't have to 100% copy ITF here

solarkraft/src/io/itf.ts
Kukovec: Oh, actually, considering the other response I made, it should be this line that stays uncommented and the line below removed. Either way, do you have a preference for one of the two approaches to dates?

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
thpani: Why do we encode a `bytes[]` as bits in the first place?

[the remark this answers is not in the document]
Kukovec: because I'm illiterate apparently

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
thpani: Then just use `.toUTCString()`?

Using i18n, and in particular `en-GB` does not make sense to me.

thpani:
@Kukovec your last commit deleted ~2kLOC from the npm lockfile.
I doubt that's intended?

Kukovec:
> @Kukovec your last commit deleted ~2kLOC from the npm lockfile. I doubt that's intended?

I have not touched it beyond what `npm run test`, `npm run lint` and `git commit` do, I have no why it's like that.

konnov:
> > @Kukovec your last commit deleted ~2kLOC from the npm lockfile. I doubt that's intended?
> 
> I have not touched it beyond what `npm run test`, `npm run lint` and `git commit` do, I have no why it's like that.

Oh, that's indeed strange. It looks like all of `yargs` and `eslint` is about to be removed from `package-lock.json`. There are plenty of packages I have no idea about though.

Kukovec:
I just took the `package-lock` that was on remote main, still no idea what caused the deletion.

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
konnov: Is this question resolved? I am not sure I understand the last comment.

solarkraft/src/state/value.ts
```diff
@@ -16,7 +16,100 @@ export type Value = {
...
+        case 'i64':
+        case 'u128':
+        case 'i128':
+        case 'symb':
+        case 'addr':
+            return v.type
+        case 'arr':
+            // As a type, `Bytes` is atomic, even though it is an array. We avoid `arr`, as it could imply
+            // a generic array type, which does not exist.
+            return 'Bytes'
```
konnov, line 58: I do not understand what it means for a type to be atomic. Atomicity means that a value is updated in a single machine instruction, or it is updated in a way such that another thread/process cannot interfere with an update. Do you mean that a value of type `Bytes` is opaque in the sense that it cannot be decomposed into smaller pieces? I am still not sure about the second sentence. What is a generic array type that does not exist?

solarkraft/src/state/value.ts
konnov: I have added a comment on that. I think it would improve our understanding of the issue if you added a link to the original description of this datatypes in Stellar or Soroban SDK?

solarkraft/src/io/itf.ts
konnov: Idk. Perhaps, using the UTC date would be more consistent. I don't have any strong preference.

konnov (approved):
I have left a few comments on the unresolved issues. I think we should resolve as much as we can asap and merge this PR.

solarkraft/src/io/itf.ts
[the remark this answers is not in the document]
Kukovec: Yes, I fixed it. `bytes = number` now, and the validations check for `2^8`

solarkraft/src/state/value.ts
Kukovec: Atomic as in the smallest type-term (or in this context, not a generic collection). E.g. `Array[T]` would be a generic array, which `Bytes` is not. There is `Vec`, which is a dynamic, non-fixed sized collection, but there are no arrays in the Rust sense.

Kukovec: I'll rewrite this a bit.

Issue #38, opened 2024-04-29
Read state from fetcher in monitor executor


thpani:
The monitor executor currently stubs out reading ITF.

- [x] #42
- [ ] #43

PR #39, opened 2024-04-29
Programmatically instrument TLA+ monitor

thpani:
Add a few helper functions to programmatically instrument a TLA+ monitor in [Apalache JSON IR](https://apalache.informal.systems/docs/adr/005adr-json.html)

This deliberately does not yet read ITF; that's left for #38.


solarkraft/src/verify.ts
```diff
@@ -15,6 +11,168 @@ import path from 'node:path'
...
+export function instrumentMonitor(
+    monitor: any,
+    state: State,
+    tx: Transaction
+): any {
+    // Add a special variable `last_error` that tracks error messages of failed transactions
+    state.push({ name: 'last_error', type: 'TlaStr', value: '' })
+
+    // Declaration of  "Init" (according to `state`)
+    const tlaInit = tlaJsonOperDecl__Conjunction(
```
konnov, line 41:
```suggestion
    const tlaInit = tlaJsonOperDecl__And(
```

It would be just a shorter name

solarkraft/src/verify.ts
```diff
@@ -15,6 +11,168 @@ import path from 'node:path'
...
+        isRecursive: false,
+        body: {
+            type: 'Untyped',
+            kind: 'OperEx',
+            oper: 'AND',
+            args: conjuncts,
+        },
+    }
+}
+
```
konnov, line 102: I know that we are still in a prototyping, but would it be good to extract this function and the ones below into a separate module?

solarkraft/test/unit/verify.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        const instrumented = instrumentMonitor(monitor, state, tx)
+        const expected = {
+            modules: [
+                {
+                    declarations: [
+                        {
+                            kind: 'TlaOperDecl',
+                            name: 'Init',
+                            type: 'Untyped',
+                            formalParams: [],
```
konnov, line 30: These JSONs are usually exploding pretty fast in tests. Would it make sense to save the expected output into a file?

solarkraft/src/verify.ts
thpani: Done 9448d16

solarkraft/test/unit/verify.test.ts
thpani: Hmmm, I don't want to read from disk in a unit test, but I've moved it to a separate module: [76d8ce1](https://github.com/freespek/solarkraft/pull/39/commits/76d8ce12f3fb15c2d1e0e5dec6be29b445f4d38b)

solarkraft/src/verify.ts
thpani: Also done: 0c9a2b1

PR #40, opened 2024-04-29
Run Apalache on instrumented monitor

thpani:
Runs Apalache on instrumented monitor:

1. Runs `apalache typecheck --output=xyz.json` to get monitor in JSON
2. Instruments monitor with state/tx data (see #39)
3. Runs `apalache check` on instrumented monitor

Reading ITF is left for #38.


konnov (approved):
This makes sense to me

PR #41, opened 2024-04-29
Add tlaplus VSCode plugin to dev container

thpani:
Add the [tlaplus VSCode plugin](https://marketplace.visualstudio.com/items?itemName=alygin.vscode-tlaplus) to the dev container spec.

This will install the plugin automatically when opening the dev container.


konnov (approved):
Nice!

PR #45, opened 2024-05-06
Read ITF in monitor executor

### LINKED ISSUE #42: Read ITF in monitor executor

thpani:
Remove hardcoded state in monitor executor and instead read state from ITF.

Introduces a transitional `--state` argument to `verify` that we can replace with convention when the fetcher writes ITF.

Closes #42


konnov (approved):
This make sense to me as an intermediate step to get something working!

PR #52, opened 2024-05-15
Add a script to populate the Setter contract

konnov:
This PR adds a bash script for populating the `Setter` contract with data, so we can use it in our integration tests.


andrey-kuprianov (commented):
LGTM

konnov:
> LGTM. I would only add that `alice` needs to be created as a prerequisite

Fixed that. Please approve.

thpani (approved):

PR #53, opened 2024-05-15
documenting interactions with Stellar SDK

konnov:
This PR contains a Markdown document that captures the knowledge about decoding Stellar ledgers, so we call know how it works


thpani (approved):
🙏

Exactly what we need

PR #55, opened 2024-05-16
the first working version of transaction fetcher

konnov:
This PR presents a working implementation of transaction decoder. The core of this contribution is:

 - [x] `solarkraft/src/fetcher/callDecoder.ts` extracts Soroban contract calls from a Horizon API.
 - [x] `solarkraft/src/fetcher/storage.ts` is the basic storage data structure that is returned by `callDecoder`.
 - [x] `solarkraft/test/integration/callDecoder.test.ts` is an integration test that fetches data from the testnet and compares the results to the expected outputs.
 - [ ] `solarkraft/src/fetch.ts` is work-in-progress code, ignore it.

The integration test is fragile as the test data will disappear after 1 month or when the testnet is reset. We have to find a more stable solution. However, for the development purposes, it should work.

### DIFF
--- a/ContractExamples/contracts/setter/src/lib.rs
+++ b/ContractExamples/contracts/setter/src/lib.rs
@@ -64,6 +64,9 @@ impl SetterContract {
         env.storage().instance().set(&MY_BOOL, &v);
         log!(&env, "myBool: {}", v);
         // bump the lifetime
+        // https://developers.stellar.org/docs/learn/smart-contract-internals/state-archival
+        //
+        // The TTL is extended to `max_ttl` when field TTL goes below `max_ttl` - 100
         env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
--- a/ContractExamples/contracts/setter/src/lib.rs
+++ b/ContractExamples/contracts/setter/src/lib.rs
@@ -64,7 +64,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_BOOL, &v);
         log!(&env, "myBool: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -78,7 +78,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_U32, &v);
         log!(&env, "myU32: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -92,7 +92,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_I32, &v);
         log!(&env, "myI32: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -106,7 +106,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_U64, &v);
         log!(&env, "myU64: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -120,7 +120,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_I64, &v);
         log!(&env, "myi64: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -134,7 +134,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_U128, &v);
         log!(&env, "myU128: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -148,7 +148,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_I128, &v);
         log!(&env, "myi128: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
@@ -162,7 +162,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_SYM, &v);
         log!(&env, "mySym: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
         // return the old value to the caller
         old
     }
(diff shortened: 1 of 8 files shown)


solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,176 @@
...
+import { Maybe, just, none } from '@sweet-monads/maybe'
+import { OrderedMap } from 'immutable'
+
+/**
+ * Decode a contract call from a Horizon operation.
+ * @param op an operation
+ * @param matcher a quick matcher over the contractId to avoid expensive deserialization
+ */
+export async function extractContractCall(
+    op: any,
```
Kukovec, line 23: Is there no way to be more precise here?

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,177 @@
...
+        // is it the right way to extract the height?
+        const height = tx.ledger_attr
+
+        // The operation itself does not give us the ledger updates.
+        // Deserialize tx.result_meta_xdr.
+        const meta = sdk.xdr.TransactionMeta.fromXDR(
+            tx.result_meta_xdr,
+            'base64'
+        )
+        if (meta.switch() !== 3) {
```
Kukovec, line 59: Can you add a comment explaining what `3` represents here?

ContractExamples/contracts/setter/src/lib.rs
```diff
@@ -64,7 +64,7 @@ impl SetterContract {
         env.storage().instance().set(&MY_BOOL, &v);
         log!(&env, "myBool: {}", v);
         // bump the lifetime
-        env.storage().instance().extend_ttl(50, 100);
+        env.storage().instance().extend_ttl(env.storage().max_ttl() - 100, env.storage().max_ttl());
```
thpani, line 67: Why do we subtract 100?

solarkraft/src/fetcher/storage.ts
```diff
@@ -0,0 +1,60 @@
...
+     * The arguments passed to the method, in native JS.
+     */
+    methodArgs: any[]
+    /**
+     * The return value in native JS.
+     */
+    returnValue: any
+    /**
+     * Ordered mapping from field names to their native values (JS).
+     * This mapping contains values only for the fields that have been set
```
thpani, line 50: So if I want all values after the tx, I have to merge `oldFields` and `fields`?

`updatedFields` may be a better name here.

solarkraft/src/fetcher/callDecoder.ts
konnov: I am not actually sure. Stellar SDK has [Operation](https://stellar.github.io/js-stellar-sdk/js-stellar-base_src_operation.js.html), but it does not look like what Horizon API is returning in Stellar SDK. Some of the methods return `sdk.Horizon.ServerApi.OperationRecord` but it's not exactly compatible with what I need.

thpani (commented):
Preliminary comments, I'm still looking at `callDecoder.ts`

ContractExamples/contracts/setter/src/lib.rs
Kukovec: +1, I had a similar question, but just assumed I was OOTL

konnov: it's a threshold of when the TTL is to be updated. I am not exactly sure how it works, so it's a bit arbitrary.

solarkraft/src/fetcher/storage.ts
konnov: No, actually, `fields` contains all fields stored so far. It's basically the new state.

konnov: Of course, if a contract removes a field from the storage, then `fields` will have fewer fields.

ContractExamples/contracts/setter/src/lib.rs
thpani: Makes sense. Could you add a brief comment? It looks a bit arbitrary.

solarkraft/src/fetcher/storage.ts
konnov: I have updated the comments a bit

thpani: Then I dont understand the comment (`contains values only for the fields that have been set by transactions`). Which fields aren't  set then?

solarkraft/src/fetcher/callDecoder.ts
konnov: Done

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @param op an operation
+ * @param matcher a quick matcher over the contractId to avoid expensive deserialization
+ */
+export async function extractContractCall(
+    op: any,
+    matcher: (contractId: string) => boolean
+): Promise<Maybe<ContractCallEntry>> {
+    // https://developers.stellar.org/network/horizon/api-reference/resources/operations/object
+    if (op.function === 'HostFunctionTypeHostFunctionTypeInvokeContract') {
+        // This operation represents a smart contract call.
```
thpani, line 28: Can't you just return `none` if `!==`.
Bit hard to follow controlflow here with the indentation.

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,178 @@
...
+        if (!matcher(contractId)) {
+            // return quickly to avoid additional calls and deserialization
+            return none()
+        }
+
+        // continue with extracting transaction data and ledger updates
+        const method = params[1]
+        const methodArgs = params.slice(2)
+
+        const tx = await op.transaction()
```
thpani, line 49: This is weird. Is it really async?

ContractExamples/contracts/setter/src/lib.rs
konnov: Added a comment. Actually, TTL is extended when it goes below the threshold, in this case, `max_ttl - 100`.

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,178 @@
...
+        // is it the right way to extract the height?
```
thpani, line 50: Is this a TODO?

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -0,0 +1,178 @@
...
+        )
+        // extract transaction metadata for version 3
+        if (meta.switch() !== 3) {
+            // Is it possible to have older transaction metadata with Soroban?
+            // TODO: make it future-proof
+            console.error(
+                `Expected TransactionMetaV3, found ${meta.switch()}: txHash = ${txHash}`
+            )
+            return none()
+        } else {
```
thpani, line 67: Also here, I don't think you need to nest inside else, since you return above?

thpani (approved):
Remaining comments, rest lgtm

ContractExamples/contracts/setter/src/lib.rs
konnov: Here is the fresh version. It's all modulo my understanding of how Soroban stores the storage in Stellar:

```js
    /**
     * Ordered mapping from field names to their native values (JS).
     * This mapping contains values only for the fields that have been created
     * or updated by a transaction in the past. It may happen that `fields` contains
     * fewer fields than `oldFields`, when the contract deletes some fields
     * from the storage. Also, `fields` may become empty when the storage
     * goes over TTL.
     */
```

solarkraft/src/fetcher/storage.ts
konnov: Here is the fresh version. It's all modulo my understanding of how Soroban stores the storage in Stellar:

```js
    /**
     * Ordered mapping from field names to their native values (JS).
     * This mapping contains values only for the fields that have been created
     * or updated by a transaction in the past. It may happen that `fields` contains
     * fewer fields than `oldFields`, when the contract deletes some fields
     * from the storage. Also, `fields` may become empty when the storage
     * goes over TTL.
     */
```

solarkraft/src/fetcher/callDecoder.ts
konnov: yeah, it looks async. Basically, SDK makes another call to the server.

solarkraft/src/fetcher/callDecoder.ts
konnov: I mean, yeah. I hope someone more knowledgeable would read it tells me to fix it :)

konnov: added `TODO:` for the future

solarkraft/src/fetcher/callDecoder.ts
konnov: yeah, good catch. It grew too large organically. Pruned it

solarkraft/src/fetcher/storage.ts
thpani: Oh, you meant values that ever have been set! Understood, lg

konnov:
I will merge this one to continue with the storage.

PR #58, opened 2024-05-16
Monitor specs for Timelock using reverse reasoning

### LINKED ISSUE #37: Investigate reverse reasoning for monitors
What we have currently as a single running example at [Timelock monitors](https://github.com/freespek/solarkraft/tree/main/doc/scf24/example) can be characterized as _direct reasoning_:

> if so-and-so cause happened (method invoked), then such-and-such effects (state changes, events) should happen

Another example of a direct reasoning is [ERC-20 symbolic properties](https://github.com/lindy-labs/solidity_properties/blob/main/PROPERTIES.md#erc20). 

The problem I see with the direct approach is that though sound, the properties/monitors for it are much more verbose than the contracts themselves.

What I want to investigate can be called a _reversed reasoning_:

> If so-and-so effects happened (state changes, events), then such-and-such causes should have happened.

Not sure yet, but this might turn out to be much less verbose than the the direct one, especially for runtime monitoring. On the one hand, properties that assert that smth. should revert are vacuously true, because no effects are present (you don't need to react, if nothing changed). On the other hand, the remaining causes (methods) could be grouped per effect they make, making the monitor specification more concise.

My immediate plan is to write an alternative representation for the Timelock monitors above, using the reversed approach, and compare it to the direct spec.

andrey-kuprianov:
TLA+ specs modeled after https://github.com/stellar/soroban-examples/blob/v20.0.0/timelock/src/lib.rs

closes #37.

The TLA+ specs are complete in the sense that the instrumented version passes `apalache-mc check` as expected.


doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. If any of `MustFail_i` conditions fire, check that method invocation reverts (otherwise, issue a warning / revert if configured to do so)
+2. If none of `MustFail_i` conditions fired, but method invocation reverted, issue a warning (incomplete spec)
+3. If none of  `MustFail_i` fired, and one of `MustSucceed_i` conditions fired, check that method invocation succeeds (otherwise, issue a warning)
+3. If none of  `MustFail_i` fired, and none of `MustSucceed_i` conditions fired, but method invocation succeeded, issue a warning of an incomplete spec (or revert if configured to do so)
+4. If method invocation succeeds, check that all of `MustAchieve_i` conditions hold on the pre- and post-states of the method invocation (otherwise, issue a warning / revert if configured to do so)
+
+
+Notice that above we apply _or_ as default connector for preconditions (`MustFail_i` / `MustSucceed_i`), and we apply _and_ as default connector for effects (`MustAchieve_i`). Thus, you may split preconditions/effects into separate predicates at your convenience, thus avoiding complicated logical structure inside predicates themselves.
```
Kukovec, line 48: You say "default", is there ever a case where this wouldn't be so?

doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+Now, is the approach described above enough to achieve the goals (safety, liveness, completeness) stated above? Think about it for a sec, before clicking on the our answer below.
```
Kukovec, line 50:
```suggestion
```
Not sure this really needs to be in the docs, this feels more like how you'd write up homework for students.

doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+\* This property would not be originally conceived 
+\* if looking solely at the timelock contract source
+MustFail_deposit_NotEnoughBalance == 
+    token.balance(from) < amount
+
+\* Balance is externally relevant; initialized flag is not
+\* What matters for users is that balance is not overwritten
+MustFail_deposit_AlreadyInitialized == 
+    exists(balance)
```
Kukovec, line 124: What is this `exists`? I can't seem to find a definition or reference.

doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+MustSucceed_deposit_Default = TRUE
+
+MustAchieve_deposit_BalanceRecordCreated ==
+    exists(balance')
+
+MustAchieve_deposit_BalanceRecordCorrect ==
+    /\ balance'.token = token
+    /\ balance'.amount = amount
+    /\ balance'.time_bound = time_bound
+    /\ balance'.claimants = claimants
```
Kukovec, line 137: From what I can find, `balance` is not a record type: https://github.com/freespek/solarkraft/blob/32f48e5f26e4cce76df4703a1b29fd1752cd1677/doc/scf24/example/timelock_mon2.tla

I know you say this spec is not valid TLA+, but having variable declarations and their types would help a lot.

Kukovec (commented):
The overall approach idea LGTM, though I have raised a point RE mustfail/mustsucceed semantics in the comments.

Since you say the monitor spec is incomplete I haven't really looked into it much. Also, per your request, I've avoided giving proofreading-suggestions, though I would recommend doing a proofreading pass before merging once all conceptual discussions are out of the way.

doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+In the above, `Must<Fail|Succeed|Achieve>` is a prefix, which tells the monitor system how to interpret this predicate. The complete pattern for predicate names with these prefixes is as follows:
+
+```
+Must<Fail|Succeed|Achieve>_<Method>_<ConditionDescription>
+```
+
+All predicates which refer to the same `<Method>` will be grouped, to create together one _method monitor_. Interpreted formally, the monitor should do the following when `<Method>` is invoked:
+
+1. If any of `MustFail_i` conditions fire, check that method invocation reverts (otherwise, issue a warning / revert if configured to do so)
+2. If none of `MustFail_i` conditions fired, but method invocation reverted, issue a warning (incomplete spec)
```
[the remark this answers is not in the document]
andrey-kuprianov: No, you don't have to give all failure scenarios. If there is a failure case which is not in the spec, there simply will be a warning (which can be also suppressed if needed), no other consequences.

doc/case-studies/timelock/README.md
andrey-kuprianov: no, it's just an unfortunate wording:) it's fixed, to that, not default

doc/case-studies/timelock/README.md
andrey-kuprianov: yeah, sorry, the defs are missing. `exists` is supposed to mean that the key is present in the ledger. We'll need to figure out the exact mapping between TLA+ and Soroban constructs

doc/case-studies/timelock/README.md
andrey-kuprianov: For the actual semantics of what both specs are supposed to represent, look at the [Timelock contract](https://github.com/stellar/soroban-examples/blob/v20.0.0/timelock/src/lib.rs). In particular, `balance` is an entry in the ledger with the [Balance key](https://github.com/stellar/soroban-examples/blob/186443aab0bf8b7c2673428c38708bf38cb772ab/timelock/src/lib.rs#L15), which points to the [ClaimableBalance struct](https://github.com/stellar/soroban-examples/blob/186443aab0bf8b7c2673428c38708bf38cb772ab/timelock/src/lib.rs#L34-L39) when the key is present.

doc/case-studies/timelock/README.md
```diff
@@ -0,0 +1,242 @@
...
+3. If none of  `MustFail_i` fired, and one of `MustSucceed_i` conditions fired, check that method invocation succeeds (otherwise, issue a warning)
+3. If none of  `MustFail_i` fired, and none of `MustSucceed_i` conditions fired, but method invocation succeeded, issue a warning of an incomplete spec (or revert if configured to do so)
```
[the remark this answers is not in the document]
andrey-kuprianov: The same. The idea is that the monitor is configurable: it may either be completely silent in some cases, or issue a simple warning, or in most severe cases interfere and abort a transaction. 

In the first stage of the project it will be at most a warning. But the ultimate goal is to provide the really protective solution for the live blockchain.

doc/case-studies/timelock/README.md
andrey-kuprianov: I also agree that the defs of the structs and everything should be present eventually. The idea now is to evaluate the approach as a whole, and to fill in the missing pieces later. 

Not directly related to this, but tangentially: the idea is that eventually the definitions wrt. the concrete contract could be auto-generated from the contract source code, so we won't have to write them by hand. But that's also not in the current stage, but in a subsequent one.

konnov:
This looks like a writeup for a blogpost. I was trying to understand how the structure imposed here would fit into the standard pre/post-conditions approach. I guess, it imposes some kind of specific shape on the monitors.

konnov:
Since we have to run `solarkraft verify` against the timelock transactions, how shall we do that? Perhaps, add notes on that?

andrey-kuprianov:
> Since we have to run `solarkraft verify` against the timelock transactions, how shall we do that? Perhaps, add notes on that?

That's a good point. Insofar it has been in isolation; I need to make sure it fits well with the rest. Will do that.

thpani (commented):
In general, I like the idea. This would make for an interesting blogpost, for sure.

In practice, I'm concerned that this will not be feasible in this round:
Remember that Apalache processes invariants into verification conditions and actions into transitions, without giving much guarantees about their correspondence. The quantifications you use here (`none/any/all of X fire`) will be hard to check, or at least require multiple instrumented specs and model-checker calls, which will become quite costly.

andrey-kuprianov:
> The quantifications you use here (`none/any/all of X fire`) will be hard to check, or at least require multiple instrumented specs and model-checker calls, which will become quite costly.

Also a good point: I of course need to make sure that this fits well with the Apalache MC. Otoh, we could say that this is the general approach we use for structuring the monitor specs, and then in this phase cut some corners, e.g. perform certain translation steps manually, and automate them in the subsequent phases.

I will investigate.

Kukovec:
I've noticed this PR seems to be creeping up on 1k LOC, would it be possible to split this into smaller PRs, which are easier to digest?

doc/case-studies/timelock/env.tla
```diff
@@ -0,0 +1,83 @@
...
+    @typeAlias: dataKey = Str;
+*)
+typedefs == TRUE
+
+VARIABLES
+    (*  Soroban's environment 
+        @type: {
+            current_contract_address: $address,
+            ledger_timestamp: Int
+        };
```
Kukovec, line 22:
```suggestion
        @type: {
            current_contract_address: $address,
            timestamp: Int,
            sequence: Int,
            invoked_function_name: Str
        };
```
I'd recommend making this the `env` type alias to use for `env`-method parameters.

doc/case-studies/timelock/env.tla
```diff
@@ -0,0 +1,83 @@
...
+
+EXTENDS Integers, Sequences, Apalache
+
+(*
+    @typeAlias: address = Str;
+    @typeAlias: dataKey = Str;
+*)
+typedefs == TRUE
+
+VARIABLES
```
Kukovec, line 17: There's a broader issue here with how this is defined: If `env`, `tx` and `instance_strorage` are to be defined as TLA state variables, then instrumentation would have to instantiate them all, which is not what happens (see [here](https://github.com/freespek/solarkraft/blob/81cf373ef8d1523e4458e49bf447cfcf10e726f6/solarkraft/src/instrument.ts#L23-L69))
The `env` gets passed as the 1st function argument, the data in `tx` is not instrumented at the moment, and neither is `instance_storage`, though the latter would probably be the easiest to simply add in.

doc/case-studies/timelock/timelock.tla
```diff
@@ -0,0 +1,30 @@
...
+ * Modeled from https://github.com/stellar/soroban-examples/blob/v20.0.0/timelock/src/lib.rs
+ *
+ * Andrey Kuprianov, 2024
+ *)
+---- MODULE timelock ----
+
+EXTENDS env, token
+
+(*
+    @typeAlias: timeBoundKind = Str;
```
Kukovec, line 13: I'd personally introduce an uninterpreted type here, e.g. `"TBK"` or even `"TIME_BOUND_KIND"`, but it's not blocking

doc/case-studies/timelock/claim.tla
```diff
@@ -0,0 +1,108 @@
...
+
+\* @type: ($claimArgs) => Bool;
+MustFail_claim_NotClaimant(args) == 
+    \A i \in DOMAIN Balance.claimants: 
+        Balance.claimants[i] /= args.claimant
+
+\* One success condition: correctly claimed before time bound
+MustPass_claim_BeforeTimeBound(args) ==
+    /\ Balance.time_bound.kind = "Before" 
+    /\ env.ledger_timestamp <= Balance.time_bound.timestamp
```
Kukovec, line 33: None of this is blocking, but from how the source code is structured:
https://github.com/yuzurush/soroban-contracts-101/blob/4f64753310b26c7853de1c9efcc0f6f2b349c15c/timelock/src/lib.rs#L104-L106
I see this (and the one below) conceptually as a `MustFail_...` predicate, since it panics if the condition doesn't hold, and executes a bunch of code afterwards if it does. So in my view:
```suggestion
MustFail_claim_BeforeTimeBound(args) ==
    /\ Balance.time_bound.kind = "Before" 
    /\ env.ledger_timestamp > Balance.time_bound.timestamp
```

doc/case-studies/timelock/token.tla
```diff
@@ -0,0 +1,43 @@
...
+
+---- MODULE token ----
+
+EXTENDS env
+
+VARIABLES
+    (*  
+        @type: $address -> $address -> Int;
+    *)
+    token_balances
```
Kukovec, line 16: Similar to my comment in `env`, be conservative with `VARIABLES`, because they'd have to be instrumented, and I don't think this one is instrumented anywhere. Do we know how instrumentation behaves, repectively should behave, w.r.t. `soroban_sdk::contractimport!`?

doc/case-studies/timelock/token.tla
```diff
@@ -0,0 +1,43 @@
...
+
+
+\* @type: ($address, $address) => Int;
+token_balance(token, id) ==
+    IF token \in DOMAIN token_balances THEN
+        IF id \in DOMAIN token_balances[token] THEN
+            token_balances[token][id]
+        ELSE 0
+    ELSE 0
```
Kukovec, line 25:
```suggestion
    IF (token \in DOMAIN token_balances /\ id \in DOMAIN token_balances[token] )THEN
            token_balances[token][id]
    ELSE 0
```
Minor backend optimization, each ITE is a `pick`

doc/case-studies/timelock/token.tla
```diff
@@ -0,0 +1,43 @@
...
+
+\* @type: ($address, $address) => Int;
+next_token_balance(token, id) ==
+    IF token \in DOMAIN token_balances' THEN
+        IF id \in DOMAIN token_balances'[token] THEN
+            token_balances'[token][id]
+        ELSE 0
+    ELSE 0
```
Kukovec, line 33: See above

doc/case-studies/timelock/claim.tla
```diff
@@ -0,0 +1,108 @@
...
+MustHold_claim(args) ==
+    /\ MustHold_claim_BalanceRecordRemoved(args)
+    \* Checking of the condition(s) below is not yet supported 
+    \* /\ MustHold_claim_TokenTransferred(args)
+
+
+\* Monitor invariants to be checked
+\* (encode the expected interpretation of monitor predicates)
+Inv_MustFail_claim(args) ==
+    (   /\ tx.method_name = "claim"
```
Kukovec, line 78: This is redundant, since the `claim(...)` operator only appears in the instrumented spec when the method name is `"claim"`

Kukovec (commented):
P1: `claim`, `env`, `monitor`, `token`

thpani (commented):
It's clear that this has been developed out of sync with the rest of the system.
I don't know what to do at this state, except to merge it and then try to incrementally make it work, which will be much harder than just incrementally adding working pieces from the beginning.

I have not looked at the tests. Could you add a shell script that executes them so we have some kind of QA here?

Below are some comments that I would like to see addressed before we merge this and take further steps.

Kukovec (approved):
Per our discussion on Friday, I'm going to merge this as-is, and use it as a reference point to write a monitor which is syntactically more aligned with our current instrumentation strategy.

PR #59, opened 2024-05-16
add a simple implementation of ADR002

konnov:
This PR implements ADR002.

### DIFF
--- a/solarkraft/src/fetcher/storage.ts
+++ b/solarkraft/src/fetcher/storage.ts
@@ -9,7 +9,12 @@
  * Igor Konnov, 2024.
  */
 
+import JSONbigint from 'json-bigint'
 import { OrderedMap } from 'immutable'
+import { join } from 'node:path/posix'
+import { mkdirSync, readFileSync, writeFileSync } from 'node:fs'
+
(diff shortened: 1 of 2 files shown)


solarkraft/src/fetcher/storage.ts
```diff
@@ -62,3 +67,68 @@ export interface ContractCallEntry {
...
+    return {
+        height: loaded.height as number,
+        contractId: loaded.contractId as string,
+        txHash: loaded.txHash as string,
+        method: loaded.method as string,
+        methodArgs: loaded.methodArgs as any[],
+        returnValue: loaded.returnValue,
+        fields: OrderedMap<string, any>(loaded.fields),
+        oldFields: OrderedMap<string, any>(loaded.oldFields),
+    }
```
Kukovec, line 107: I think I'd prefer to have a validation method, which checks that `loaded` has the correct shape, and then returns the value cast to the correct type with given field types, instead of just blind-casting each field.

konnov: Please feel free to implement one!

Kukovec (approved):
Not going to block with the above, LGTM generally.

PR #60, opened 2024-05-17
working fetcher CLI

konnov:
This PR adds relatively small changes on top of #55. The fetcher is fetching the transactions and saving them in the storage. For example:

```sh
$ solarkraft fetch --rpc https://horizon-testnet.stellar.org --id CC22QGTOUMERDNIYN7TPNX3V6EMPHQXVSRR3XY56EADF7YTFISD2ROND --height 1638368
Target contract: CC22QGTOUMERDNIYN7TPNX3V6EMPHQXVSRR3XY56EADF7YTFISD2ROND...
Last cached height: 1638598
Fetching fresh transactions from: https://horizon-testnet.stellar.org...
Fetching the ledger for 1638368
Fetching transactions indefinitely. Close with Ctrl-C.
+ save: 1638369
+ save: 1638371
+ save: 1638370
+ save: 1638372
+ save: 1638373
+ save: 1638374
+ save: 1638375
+ save: 1638376
+ save: 1638378
+ save: 1638377
+ save: 1638379
+ save: 1638380
+ save: 1638381
+ save: 1638382
+ save: 1638383
= at: 1638412
= at: 1638456
```

I will add an integration test for the above later today.

### DIFF
--- a/solarkraft/src/fetch.ts
+++ b/solarkraft/src/fetch.ts
@@ -11,7 +11,11 @@
 
 import { Horizon } from '@stellar/stellar-sdk'
 import { extractContractCall } from './fetcher/callDecoder.js'
-import { loadFetcherState, saveContractCallEntry, saveFetcherState } from './fetcher/storage.js'
+import {
+    loadFetcherState,
+    saveContractCallEntry,
+    saveFetcherState,
+} from './fetcher/storage.js'
 
 // how often to query for the latest synchronized height
 const HEIGHT_FETCHING_PERIOD = 100
@@ -40,7 +44,7 @@ export async function fetch(args: any) {
     }
 
     console.log(`Fetching fresh transactions from: ${args.rpc}...`)
-    
+
     console.log(`Fetching the ledger for ${lastHeight}`)
     const response = await server.ledgers().ledger(lastHeight).call()
     const startCursor = (response as any).paging_token
@@ -65,8 +69,10 @@ export async function fetch(args: any) {
         .stream({
             onmessage: async (msg: any) => {
                 if (msg.transaction_successful) {
-                    const callEntryMaybe =
-                        await extractContractCall(msg, (id) => contractId === id)
+                    const callEntryMaybe = await extractContractCall(
+                        msg,
+                        (id) => contractId === id
+                    )
                     if (callEntryMaybe.isJust()) {
                         const entry = callEntryMaybe.value
                         console.log(`+ save: ${entry.height}`)
@@ -83,7 +89,13 @@ export async function fetch(args: any) {
                     console.log(`= at: ${lastHeight}`)
                     // load and save the state, other fetchers may work concurrently
                     fetcherState = loadFetcherState(args.home)
-                    fetcherState = { ...fetcherState, heights: fetcherState.heights.set(contractId, lastHeight) }
+                    fetcherState = {
+                        ...fetcherState,
+                        heights: fetcherState.heights.set(
+                            contractId,
+                            lastHeight
+                        ),
+                    }
                     saveFetcherState(args.home, fetcherState)
                 }
             },
--- a/solarkraft/src/fetch.ts
+++ b/solarkraft/src/fetch.ts
@@ -35,8 +35,8 @@ export async function fetch(args: any) {
     console.log(`Last cached height: ${cachedHeight}`)
     if (args.height < 0) {
         // how to fetch the latest height?
-        console.log('not implemented yet, starting with 0')
-        lastHeight = 0
+        console.log(`not implemented yet, starting with ${cachedHeight}`)
+        lastHeight = cachedHeight
     } else if (args.height === 0) {
         lastHeight = cachedHeight
     } else {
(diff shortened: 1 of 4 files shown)


solarkraft/src/fetcher/storage.ts
```diff
@@ -68,13 +68,33 @@ export interface ContractCallEntry {
...
 
+/**
+ * Serializable fetcher state.
+ */
+export interface FetcherState {
+    /**
+     * For every contract id, store the ledger height,
+     * up to which the transactions were fetched.
+     */
+    heights: OrderedMap<string, number>
```
thpani, line 79: Should we use bigint as the value?

solarkraft/src/fetcher/storage.ts
```diff
@@ -107,6 +127,42 @@ export function loadContractCallEntry(filename: string): ContractCallEntry {
...
+ * Load fetcher state from the storage.
+ * @param root the storage root directory
+ * @returns the loaded state
+ */
+export function loadFetcherState(home: string): FetcherState {
+    const filename = getFetcherStateFilename(home)
+    if (!existsSync(filename)) {
+        // just return an empty map
+        return {
+            heights: OrderedMap<string, any>()
```
thpani, line 140: Why not

```suggestion
            heights: OrderedMap<string, number>()
```

solarkraft/src/fetcher/storage.ts
```diff
@@ -107,6 +127,42 @@ export function loadContractCallEntry(filename: string): ContractCallEntry {
...
+        }
+    } else {
+        const contents = readFileSync(filename)
+        const loaded = JSONbig.parse(contents)
+        return {
+            heights: OrderedMap<string, any>(loaded.heights),
```
thpani, line 146: Why not

```suggestion
            heights: OrderedMap<string, number>(loaded.heights),
```

solarkraft/src/fetcher/storage.ts
```diff
@@ -119,6 +175,18 @@ function getEntryFilename(root: string, entry: ContractCallEntry) {
...
 
+/**
+ * Get the filename for a contract call entry. Create the parent directory, if required.
+ *
+ * @param root storage root
+ * @param entry call entry
+ * @returns the filename
+ */
+function getFetcherStateFilename(root: string) {
+    mkdirSync(root, { recursive: true })
```
thpani, line 186: Not really obvious that this has side-effects from the function name.

Can we just call `mkdirSync` in `saveFetcherState` instead of here?

solarkraft/src/fetch.ts
```diff
@@ -10,64 +10,90 @@
...
-    */
-    const startHeight = args.height
+    let fetcherState = loadFetcherState(args.home)
+    const cachedHeight = fetcherState.heights.get(contractId, 1)
+    let lastHeight = args.height
+    console.log(`Last cached height: ${cachedHeight}`)
+    if (args.height < 0) {
+        // how to fetch the latest height?
+        console.log('not implemented yet, starting with 0')
+        lastHeight = 0
```
thpani, line 35: Shouldn't this also start from the cached height?

solarkraft/src/fetch.ts
```diff
@@ -10,64 +10,90 @@
...
+                if (nEvents % HEIGHT_FETCHING_PERIOD === 0) {
+                    // Fetch the height of the current message.
+                    // Note that messages may come slightly out of order, so the heights are not precise.
+                    const tx = await msg.transaction()
+                    lastHeight = Math.max(lastHeight, tx.ledger_attr)
+                    console.log(`= at: ${lastHeight}`)
+                    // load and save the state, other fetchers may work concurrently
+                    fetcherState = loadFetcherState(args.home)
+                    fetcherState = { ...fetcherState, heights: fetcherState.heights.set(contractId, lastHeight) }
+                    saveFetcherState(args.home, fetcherState)
```
thpani, line 87: Don't you get a filesystem race here with concurrent fetchers? 😄 

And actually, will the streaming be sequential?

solarkraft/src/fetch.ts
```diff
@@ -10,64 +10,90 @@
...
+                        const entry = callEntryMaybe.value
+                        console.log(`+ save: ${entry.height}`)
+                        saveContractCallEntry(args.home, entry)
+                    }
+                } // else: shall we also store reverted transactions?
+
+                nEvents++
+                if (nEvents % HEIGHT_FETCHING_PERIOD === 0) {
+                    // Fetch the height of the current message.
+                    // Note that messages may come slightly out of order, so the heights are not precise.
```
thpani, line 80: Not entirely sure what this block does.

afaict, every 100 messages, it will actually fetch the tx blockheight, and updated the persisted state? May be worth a comment.

thpani (approved):
I have a few comments, but nothing blocking.

Thanks for pushing this through 🙏

solarkraft/src/fetcher/storage.ts
konnov: Yeah, I also thought about it, but Stellar SDK is using `number` so we are compatible with them: https://stellar.github.io/js-stellar-sdk/OperationCallBuilder.html#forLedger

solarkraft/src/fetcher/storage.ts
konnov: Now I am surprised that `tsc` has not reported that :)

solarkraft/src/fetch.ts
konnov: good catch, I will add a comment

solarkraft/src/fetch.ts
konnov: yeah, there is a possibility into missing filesystem updates for concurrently running fetchers. Also, for the callbacks in theory.

PR #64, opened 2024-05-17
add a simple implementation of `solarkraft list`

konnov:
This PR implements `solarkraft list`. Since our file storage is very simple, the implementation is also simple.

### DIFF
--- a/solarkraft/src/fetcher/storage.ts
+++ b/solarkraft/src/fetcher/storage.ts
@@ -154,13 +154,14 @@ export function* yieldListEntriesForContract(
     path: string
 ): Generator<ListEntry> {
     for (const dirent of readdirSync(path, { withFileTypes: true })) {
+        // match ledger heights, which are positive integers
         if (dirent.isDirectory() && /^[0-9]+$/.exec(dirent.name)) {
             // This directory may contain several transactions for the same height.
             const height = Number.parseInt(dirent.name)
             for (const ledgerDirent of readdirSync(join(path, dirent.name), {
                 withFileTypes: true,
             })) {
-                // match all storage entries
+                // match all storage entries, which may be reported in different cases
                 const matcher = /^entry-([0-9a-fA-F]+)\.json$/.exec(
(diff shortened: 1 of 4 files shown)


solarkraft/src/fetcher/storage.ts
```diff
@@ -128,6 +144,41 @@ export function loadContractCallEntry(filename: string): ContractCallEntry {
...
+ * Generate storage entries for a given contract id in a path.
+ * @param contractId contract identifier (address)
+ * @param path the path to the contract storage
+ */
+export function* yieldListEntriesForContract(
+    contractId: string,
+    path: string
+): Generator<ListEntry> {
+    for (const dirent of readdirSync(path, { withFileTypes: true })) {
+        if (dirent.isDirectory() && /^[0-9]+$/.exec(dirent.name)) {
```
Kukovec, line 157: Please add a comment explaining this regex

solarkraft/src/fetcher/storage.ts
```diff
@@ -128,6 +144,41 @@ export function loadContractCallEntry(filename: string): ContractCallEntry {
...
+            // This directory may contain several transactions for the same height.
+            const height = Number.parseInt(dirent.name)
+            for (const ledgerDirent of readdirSync(join(path, dirent.name), {
+                withFileTypes: true,
+            })) {
+                // match all storage entries
+                const matcher = /^entry-([0-9a-fA-F]+)\.json$/.exec(
```
Kukovec, line 164: Why are we using mixed-case HEX here?

Kukovec (approved):
LGTM

solarkraft/src/fetcher/storage.ts
konnov: I would be ok with lowercase, but neither Windows, nor MacOS agree with me on the capitalization of filenames

konnov: added a comment

PR #68, opened 2024-05-21
Alert contract draft

Kukovec:
First step towards #67

We should understand if this is the interface we want from the Alert contract or if we want to go in a different direction.

### DIFF
--- a/ContractExamples/contracts/alert/src/lib.rs
+++ b/ContractExamples/contracts/alert/src/lib.rs
@@ -1,0 +1,54 @@
+#![no_std]
+use soroban_sdk::{contract, contractimpl, contracttype, symbol_short, Env, Symbol, Map, String};
+
+const ALERTS: Symbol = symbol_short!("ALERTS");
+const VIOLATION: Symbol = symbol_short!("VIOLATION");
+
+#[contract]
+pub struct Alert;
+
+#[contracttype]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(u32)]
+pub enum MonitorAnalysisStatus {
+    NoViolation = 0,
+    Violation = 1,
+    Unknown = 2
+}
+
+type AlertMap = Map<String, MonitorAnalysisStatus>;
+
+/*
+    The Alert contract gets called by the monitor executor, whenever a transaction gets analyzed.
+    If the monitor detected a property violation, then the Alert contract emits a warning.
+
+    For each tx hash, for which alert was called, stores MonitorAnalysisStatus, the value of which
+    depends on whether the monitor detected a violation for that transaction.
+*/
+#[contractimpl]
+impl Alert {
+
+    pub fn emit_warning_if_violation(env: Env, tx_hash: String, status: MonitorAnalysisStatus) -> MonitorAnalysisStatus {
+        // Get the current alerts
+        let mut alerts: AlertMap = env.storage().instance().get(&ALERTS).unwrap_or(AlertMap::new(&env));
+
+        // Add to history and save
+        alerts.set(tx_hash, status);
+        env.storage().instance().set(&ALERTS, &alerts);
+
+        // The contract instance will be bumped to have a lifetime of at least 100 ledgers if the current expiration lifetime at most 50.
+        // If the lifetime is already more than 100 ledgers, this is a no-op. Otherwise,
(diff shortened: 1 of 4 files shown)


ContractExamples/contracts/alert/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        // Add to history and save
+        alerts.set(tx_hash, status);
+        env.storage().instance().set(&ALERTS, &alerts);
+
+        // The contract instance will be bumped to have a lifetime of at least 100 ledgers if the current expiration lifetime at most 50.
+        // If the lifetime is already more than 100 ledgers, this is a no-op. Otherwise,
+        // the lifetime is extended to 100 ledgers. This lifetime bump includes the contract
+        // instance itself and all entries in storage().instance().
+        env.storage().instance().extend_ttl(50, 100);
```
thpani, line 43: I think this was a hack for Igor's test contract. If you use `instance`, the storage TTL will be the contract's TTL, so I don't think it needs to be bumped.

ContractExamples/contracts/alert/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        // If there is a violation, emit an event.
+        if status == MonitorAnalysisStatus::Violation {
```
thpani, line 46: Probably we'd want every result emitted? Or different events for each outcome?

thpani (commented):
Do we want one alert contract per monitored contract?

konnov (commented):
one comment

konnov:
> Do we want one alert contract per monitored contract?

Yeah, that would be nice, but how would we discover this alert contract? We would need a repository contract then?

ContractExamples/contracts/alert/src/lib.rs
thpani: Why? Persistent gives an individual TTL to every field.

Kukovec: My understanding is that `Instance` still gives a finite TTL, so you do need to periodically extend it.

Kukovec: Actually, reading [this](https://developers.stellar.org/docs/smart-contracts/guides/storage/use-instance) makes me even more confused as to why `.extend_ttl` even exists/what it does. If "Under the hood, instance storage is exactly like persistent storage. The only difference is that anything stored in instance storage has an archival TTL that is tied to the contract instance itself. So, if a contract is live and available, the instance storage is guaranteed to be so, too" is true, you would never need to actually `extend_ttl`

thpani: Verbatim from the section you linked: `For Temporary and Persistent storage, each entry has its own TTL and must be extended individually.`

Kukovec: No, I mean what is the point of `.instance().extend_ttl(...)` given the above.

Kukovec: You're probably right that `Instance` is the best choice for this sort of contract regardless.

ContractExamples/contracts/alert/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+const VIOLATION: Symbol = symbol_short!("VIOLATION");
+
+#[contract]
+pub struct Alert;
+
+#[contracttype]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(u32)]
+pub enum MonitorAnalysisStatus {
+    NoViolation = 0,
```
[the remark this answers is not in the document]
konnov: I am actually not sure about this. We could probably get meaningful feedback from the users. The thing is that the events are actually a temporary data structure that is usually monitored by offchain processes. If we save the result of no violation in the contract storage, we are going to fill in the storage pretty fast.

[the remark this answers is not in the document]
konnov: yes

konnov:
> > Yeah, that would be nice, but how would we discover this alert contract? We would need a repository contract then?
> 
> You could do this with `N+1` contracts: a central `AlertDispatcher`, which gets called by solarkraft, and delegates further to the individual `AlertForX` contract associated with a particular monitored contract `X`

Let's keep things simple for this phase. One alert contract for everything.

PR #70, opened 2024-05-21
Fix typo in symbol

thpani:
Fix a symbol typo in the setter contract


Kukovec (approved):

konnov:
Hmm. I don't think it was a typo. There is a constraint on symbols having at most 10 bytes. Have you tried to deploy the new version?

Issue #73, opened 2024-05-21
Support custom types in monitor executor


thpani:
Add support for `enum` and `struct` types

PR #74, opened 2024-05-21
Support Map and Vec

### LINKED ISSUE #46: Support aggregate types in monitor executor
So far we only support basic ITF types in the executor, when instrumenting the monitor spec.
Extend support to composite types (sequence, tuple, record, set, map).

thpani:
Add support for Soroban `Map` and `Vec`.

As with #72, in addition to the unit tests here, I have an on-chain E2E test based on the setter contract that I'll add in a follow-up PR once all its fields/types are supported.

Closes #46 

Support for `enum` and `struct` will be a follow-up: #73


konnov (approved):
nice!

PR #75, opened 2024-05-22
Adds programmatic alert contract invocation

### LINKED ISSUE #67: Implement `Alert` contract

Kukovec:
Closes #67 

Adds a method `invokeAlert`, which submits the monitor analysis status to the alert contract.
Calling `invokeAlert` requires a known `sorobanRpcServer`, as well as an account/keypair with which to sign/submit transactions. In the test, the latter are randomly generated and funded, but we should probably hardcode them to cut down on response waiting times.

### DIFF
--- a/ContractExamples/scripts/alert-deploy.sh
+++ b/ContractExamples/scripts/alert-deploy.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# Deploy the alert contract
+#
+# Jure Kukovec, 2024
+#
+# @license
+# [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
+
+
+set -e
+
+dir=$(cd `dirname $0`; pwd -P)
+
+cd ${dir}/..
+
+NET=testnet
+(soroban network ls | grep -q $NET) || (echo "add testnet via soroban network"; exit 1)
+
+ACCOUNT=alice
+soroban keys address $ACCOUNT || (echo "add the account $ALICE via soroban keys generate"; exit 1)
+
+
+set -x
+
+soroban contract build
+soroban contract deploy --wasm target/wasm32-unknown-unknown/release/alert.wasm \
+      --source $ACCOUNT --network $NET
--- a/solarkraft/package.json
+++ b/solarkraft/package.json
@@ -30,6 +30,7 @@
     "dist/**/*"
   ],
   "scripts": {
+    "alert": "mocha --timeout 0 --loader=ts-node/esm test/alert/**/*.test.ts",
     "compile": "genversion -e src/version.ts && tsc",
     "e2e": "mocha --loader=ts-node/esm test/e2e/**/*.test.ts",
     "format": "prettier --config .prettierrc 'src/**/*.ts' 'test/**/*.ts' --write",
--- a/solarkraft/src/invokeAlert.ts
+++ b/solarkraft/src/invokeAlert.ts
@@ -1,0 +1,116 @@
+import {
+    Contract,
+    SorobanRpc,
+    TransactionBuilder,
+    Networks,
+    BASE_FEE,
+    xdr,
+    Keypair,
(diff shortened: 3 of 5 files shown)


solarkraft/package.json
```diff
@@ -30,14 +30,15 @@
     "dist/**/*"
   ],
   "scripts": {
+    "alert": "mocha --timeout 0 --loader=ts-node/esm test/alert/**/*.test.ts",
```
thpani, line 33: Why not put it under `e2e` which already exists?

ContractExamples/scripts/alert-deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+
+dir=$(cd `dirname $0`; pwd -P)
+
+cd ${dir}/..
+
+NET=testnet
+(soroban network ls | grep -q $NET) || (echo "add testnet via soroban network"; exit 1)
+
+ACCOUNT=alice
+soroban keys address $ACCOUNT || (echo "add the account $ALICE via soroban keys generate"; exit 1)
```
thpani, line 21: Where does `$ALICE` come from?

konnov: It's like in the [setter-populate.sh](https://github.com/freespek/solarkraft/blob/main/ContractExamples/scripts/setter-populate.sh). The user (or the testing environment) has to create this account.

solarkraft/package.json
```diff
@@ -30,14 +30,15 @@
...
     "compile": "genversion -e src/version.ts && tsc",
     "e2e": "mocha --loader=ts-node/esm test/e2e/**/*.test.ts",
     "format": "prettier --config .prettierrc 'src/**/*.ts' 'test/**/*.ts' --write",
     "license": "source-licenser --config-file licenser-config.yaml .",
     "lint": "eslint src test",
     "prepare": "cd .. && husky solarkraft/.husky",
     "snapshot": "git archive --format=tar.gz -o solarkraft-`git rev-parse --short HEAD`.tar.gz --prefix=solarkraft/ HEAD",
-    "test": "mocha --loader=ts-node/esm --exclude test/e2e/**/*.test.ts test/**/*.test.ts"
+    "test": "mocha --loader=ts-node/esm --exclude test/e2e/**/*.test.ts --exclude test/alert/**/*.test.ts test/**/*.test.ts"
```
konnov, line 41: I agree with @thpani, we should simply include the alert tests in e2e

solarkraft/src/invokeAlert.ts
```diff
@@ -0,0 +1,116 @@
...
+import { Api } from '@stellar/stellar-sdk/lib/soroban/api.js'
+
+/**
+ * @license
+ * [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
+ */
+
+const AlertMethodName = 'emit_and_store_violation'
+export const LocalRPC = 'https://localhost:8000/'
+export const TestnetRPC = 'https://soroban-testnet.stellar.org:443'
```
konnov, line 21: can we move both of these definitions to the test, as they are used in the test

solarkraft/package.json
Kukovec: Well, partially because it's not really end-to-end, more like mid-to-end, but more prominently, because this way it gets a separate timeout flag, since it's currently doing some back-and-forth requests each time, and so the test takes longer than all the rest (which have a default of 2s timeout)

konnov (approved):
This looks great! Please fix the comments by @thpani and myself

ContractExamples/scripts/alert-deploy.sh
thpani: What I meant is that there's no variable `$ALICE`. Only `$ACCOUNT`

solarkraft/package.json
thpani: You can set the timeout. See the e2e tests

PR #79, opened 2024-05-24
Fixes hardcoded Apalache path

### LINKED ISSUE #34: Fix hardcoded path to Apalache
Apalache is currently found via a hard-coded path that works in the devcontainer.

Make it more flexible (e.g., check in $PATH).

Kukovec:
closes #34

Instead of looking at a fixed directory, attempts to read the `$APALACHE_HOME` environment variable.

### DIFF
--- a/solarkraft/src/environment.d.ts
+++ b/solarkraft/src/environment.d.ts
@@ -1,0 +1,13 @@
+/**
+ * @license
+ * [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
+ */
+declare global {
+    namespace NodeJS {
+        interface ProcessEnv {
+            APALACHE_HOME: string
+        }
+    }
+}
+
+export {}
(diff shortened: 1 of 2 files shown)


solarkraft/src/environment.d.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+declare global {
+    namespace NodeJS {
+        interface ProcessEnv {
+            APALACHE_HOME: string
+        }
+    }
+}
+
+export {}
```
konnov, line 13: why do we need that? Is it some kind of an idiom for the case of running outside of node?

konnov (approved):
Looks good. I am just a bit puzzled about `environment.d.ts`

solarkraft/src/environment.d.ts
Kukovec: It's used to specify the `process.env` object's `APALACHE_HOME` field.

konnov: Cool. I did not know about that!

PR #80, opened 2024-05-27
Persistent verification status

### LINKED ISSUE #66: Persist verification result
Persist the verification result in the checker and report it in `list`

Kukovec:
closes #66

### DIFF
--- a/solarkraft/.gitignore
+++ b/solarkraft/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 dist
+_apalache-out
--- a/solarkraft/src/fetcher/storage.ts
+++ b/solarkraft/src/fetcher/storage.ts
@@ -28,6 +28,7 @@ const JSONbig = JSONbigint({ useNativeBigInt: true })
  */
 export type FieldsMap = OrderedMap<string, any>
 
+export type VerificationStatus = 'ok' | 'fail' | 'unverified'
 /**
  * A storage entry for a performed contract call.
  */
@@ -72,6 +73,11 @@ export interface ContractCallEntry {
      * when the storage goes over TTL.
      */
     oldFields: FieldsMap
+
+    /**
+     * Flag which tracks whether this particular entry has already been verified, and, if it was, the verification result.
+     */
+    verification?: VerificationStatus
 }
 
 /**
@@ -81,7 +87,7 @@ export interface ListEntry {
     contractId: string
     height: number
     txHash: string
-    verification: 'ok' | 'fail' | 'unverified'
+    verification: VerificationStatus
 }
 
 /**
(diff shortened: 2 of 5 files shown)


solarkraft/src/fetcher/storage.ts
```diff
@@ -167,12 +177,19 @@ export function* yieldListEntriesForContract(
...
+                    const contents = JSONbig.parse(
+                        readFileSync(filename, 'utf-8')
+                    )
+                    const status = contents['verification'] ?? 'unverified'
                     yield {
                         contractId,
                         height,
                         txHash,
-                        verification: 'unverified',
+                        verification: status,
```
thpani, line 192: I think it would be more idiomatic to just call the field `verificationStatus` and do this here:

```suggestion
                        verificationStatus,
```

Also from the UX perspective, I think just calling the field `verification` in the JSON can be misleading.

solarkraft/src/verify.ts
```diff
@@ -147,16 +154,25 @@ export function verify(args: any) {
                     console.log(
                         `${path.basename(monitorPath)}: ${errorStringOrOK}`
                     )
+                    apalacheResult
+                        .map(() => (verificationStatus = 'ok'))
```
thpani, line 158: I wouldn't expect an assignment inside the functional part. Isn't `apalacheResult` redundant after this? Why not return the result of this map/mapLeft?

solarkraft/.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 dist
+_apalache-out
+entry-0xasdf.json
```
thpani, line 4: Why is this here? Looks like a leftover?

thpani (commented):
LGTM, but I don't like the side-effects in the functional code.

Is it really necessary this way?

solarkraft/.gitignore
Kukovec: If you look at `asdf/100/entry-timelock.json` test file, even though it's called that, the contents _inside_ the file say
```
"txHash": "0xasdf",
"contractId": "0xqwer",
```
instead of
```
"txHash": "timelock",
"contractId": "asdf",
```
So if you actually run `verify` on this file, the output will be saved to a file, the name of which will be synthesized from the content, i.e. `0xqwer/100/entry-0xasdf.json`

Although I suppose if the gitignore edit bothers you, an easier soloution would probably be to change the hash and id in the original json.

thpani: > Although I suppose if the gitignore edit bothers you

It's not about what bothers me.
Renaming a misnamed file just looks like the more immediate solution.

thpani (approved):

PR #82, opened 2024-05-29
Modified instrumentation

### LINKED ISSUE #81: Support transactions where the fields change
It may be the case that the fields in `fields` and `oldFields` of a transaction do not match, because of the way data storage is handled. To avoid issues with Apalache, due to missing state-variable assignments, we should use `Gen`-based initialization to mitigate this.

Kukovec:
closes #81
This PR changes instrumentation, to the format discussed during our recent meeting. A detailed explanation of the format can be found inside `instrument.ts`.
To this end, it makes the following changes:
  - Instrumentation now constructs `Init` from `oldFields`, and `Next` from a conjunction of state-variable updates from `fields` and transition constraints from `method` and `methodArgs`.
  - Instrumentation adds `Gen(1)` initializations (in either `Init` or `Next`) to all variables declared in the monitor, but absent from the transaction data (in `oldFields` or `fields`, respectively). This allows us to avoid Apalache errors on transactions in which data storage changes.
  - Tests with expecetd instrumentation syntax have been updated to reflect the changes. No semantic changes have been made in those tests
  - New tests covering the newly-supported initialization paradigm have been added
  - `entry-timelock.json` has been fixed to comply with the transition requirements of `Claim` (`last_error' = "contract is not initialized"` whenever `~is_initialized`)


solarkraft/src/instrument.ts
```diff
@@ -6,6 +6,52 @@ import { ContractCallEntry } from './fetcher/storage.js'
  * [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
  */
 
+/**
+ * Expected monitor shape:
+ * We assume the monitors are organized in the following way:
+ *  - 1 central monitor specification `Spec`, which declares invariants relating to the pre- and post-transaction states.
+ *    No invariant delcared in this specification reasons about contract methods or their internals.
```
thpani, line 13:
```suggestion
 *    No invariant declared in this specification reasons about contract methods or their internals.
```

solarkraft/src/instrument.ts
```diff
@@ -6,6 +6,52 @@ import { ContractCallEntry } from './fetcher/storage.js'
...
+ *  - 1 method monitor per method. For each method `foo`, we expect a `foo.tla` specification, which delcares no variables
```
thpani, line 14:
```suggestion
 *  - 1 method monitor per method. For each method `foo`, we expect a `foo.tla` specification, which declares no variables
```

solarkraft/src/instrument.ts
```diff
@@ -6,6 +6,52 @@ import { ContractCallEntry } from './fetcher/storage.js'
...
+ *    and defines a top-level operator `foo(...)`, representing the collection of all method invariants.
+ *    This operator takes `k` arguments, where `k` is the number of arguments declared for the contract method `foo`,
+ *    in the same order as the contract method `foo`. The operator  parameter names should, but are not required to,
```
thpani, line 17:
```suggestion
 *    in the same order as the contract method `foo`. The operator parameter names should, but are not required to,
```

solarkraft/src/instrument.ts
```diff
@@ -18,21 +64,40 @@ export function instrumentMonitor(
 ): any {
     // Only instrument state variables that are delcared in the monitor spec
     // (otherwise we provoke type errors in Apalache Snowcat, via undeclared variables)
+    // Since `fields` and `oldFields` can have different entries, due to data storage lifetimes,
+    // we initialize all variables missing from either fields collection with Gen(1)
```
thpani, line 68: I think the first 2 lines are obsolete.
Also changed the last sentence a bit, the clarify what it means to be "missing".

```suggestion
    // Since `fields` and `oldFields` can have different entries, due to data storage lifetimes,
    // we initialize all variables that appear in the monitor specification but are 
    // missing from either fields collection, with Gen(1).
```

solarkraft/src/instrument.ts
```diff
@@ -18,21 +64,40 @@ export function instrumentMonitor(
...
+    //   - missingOldFields = [B]
     const declaredMonitorVariables = monitor.modules[0].declarations
         .filter(({ kind }) => kind == 'TlaVarDecl')
         .map(({ name }) => name)
-    const fieldsToInstrument = contractCall.oldFields.filter((value, key) =>
+    const oldFieldsToInstrument = contractCall.oldFields.filter((value, key) =>
+        declaredMonitorVariables.includes(key)
+    )
+    const missingOldFields = declaredMonitorVariables.filter(
+        (k) => !contractCall.oldFields.keySeq().contains(k)
```
thpani, line 85:
```suggestion
        k => !contractCall.oldFields.has(k)
```

solarkraft/src/instrument.ts
```diff
@@ -18,21 +64,40 @@ export function instrumentMonitor(
...
+    )
+    const fieldsToInstrument = contractCall.fields.filter((value, key) =>
         declaredMonitorVariables.includes(key)
     )
+    const missingFields = declaredMonitorVariables.filter(
+        (k) => !contractCall.fields.keySeq().contains(k)
```
thpani, line 91:
```suggestion
        k => !contractCall.fields.has(k)
```

solarkraft/src/instrument.ts
```diff
@@ -44,16 +109,43 @@ export function instrumentMonitor(
             .toArray()
     )
 
+    const oldMissing = tlaJsonAnd(
+        missingOldFields.map((name) =>
```
thpani, line 113:
```suggestion
        missingOldFields.map(name =>
```

solarkraft/src/instrument.ts
```diff
@@ -44,16 +109,43 @@ export function instrumentMonitor(
...
+            tlaJsonEq__NameEx__ValEx(name, false, GEN1)
+        )
+    )
+
+    const tlaInit = tlaJsonOperDecl__And('Init', [oldInstrumented, oldMissing])
+
     // Declaration of "Next" (according to `tx`)
```
thpani, line 120:
```suggestion
    // Declaration of "Next" (according to `contractCall.fields` and `contractCall.method` / `.methodArgs`)
```

solarkraft/src/instrument.ts
```diff
@@ -44,16 +109,43 @@ export function instrumentMonitor(
...
+                    name,
+                    true, // prime must be true in next
+                    tlaJsonOfNative(value)
+                )
+            })
+            .valueSeq()
+            .toArray()
+    )
+    const currentMissing = tlaJsonAnd(
+        missingFields.map((name) => tlaJsonEq__NameEx__ValEx(name, true, GEN1)) // prime must be true in next
```
thpani, line 138:
```suggestion
        missingFields.map((name) => tlaJsonEq__NameEx__ValEx(name, true, GEN1)) // name' = Gen(1)
```

solarkraft/src/instrument.ts
```diff
@@ -44,16 +109,43 @@ export function instrumentMonitor(
...
+
+    const currentInstrumented = tlaJsonAnd(
+        fieldsToInstrument
+            .map((value, name) => {
+                return tlaJsonEq__NameEx__ValEx(
+                    name,
+                    true, // prime must be true in next
+                    tlaJsonOfNative(value)
+                )
+            })
```
thpani, line 133:
```suggestion
            .map((value, name) => tlaJsonEq__NameEx__ValEx(
                    name,
                    true, // prime `name`
                    tlaJsonOfNative(value)
                )
            )
```

solarkraft/src/instrument.ts
```diff
@@ -135,6 +227,20 @@ export function tlaJsonOfNative(v: any): any {
     return tlaJsonValEx(tlaJsonTypeOfPrimitive(v), v)
 }
 
+/**
+ * Return an Apalache TLA+ expression of the form `conjuncts`_0 /\ ... /\ `conjuncts`_n.
+ * @param conjuncts Body conjuncts
+ * @returns The conjunction in Apalache IR JSON: https://apalache.informal.systems/docs/adr/005adr-json.html
+ */
+function tlaJsonAnd(conjuncts: any): any {
```
thpani, line 235:
```suggestion
function tlaJsonAnd(conjuncts: any[]): any {
```

solarkraft/test/unit/verify.instrumentedMonitor.ts
```diff
@@ -19,24 +19,37 @@ const instrumentedMonitor = {
...
+                                        ],
                                     },
                                 ],
                             },
+                            {
+                                kind: 'OperEx',
+                                oper: 'AND',
+                                type: 'Untyped',
+                                args: [],
+                            },
```
thpani, line 52: Ah interesting, I had not considered before that empty arrays will generate empty conjunctions.

It's correct ofc and inconsequential for now, but I'm kinda scared this will produce a mean bug if we ever generate disjunctions in the future 😅

thpani (approved):
Looks great!

I'm a bit scared about the empty conjunction (see my comment below) if the code evolves.

solarkraft/test/unit/verify.instrumentedMonitor.ts
Kukovec: That's a reasonable concern, do you think it'd be better if I merged it so that instead of 
```
and(
  and(present1, ..., presentN),
  and(missing1, ..., missingM)
)
```
it's just a flat
```
and(present1, ..., presentN, missing1, ..., missingM)
```
? It shouldn't be too complicated, since you're just merging two argument arrays.

thpani: Yeah, I thought about it, but I'm undecided...
In the end, it's doing what it's supposed to do, and `tlaAnd` does imply what happens if you pass an empty array. I think we can keep it as-is.

Kukovec: 👍

Kukovec:
@thpani just fyi, it seems like the `prettier` reverted some of your method argument parenthesis suggestions, it's not me manually reverting these

PR #83, opened 2024-05-29
Fetch and instrument block timestamp

thpani:
Fetch and instrument the block timestamp, which is available in Soroban as `env.ledger().timestamp()`.

Also modifies and moves around some e2e test assets, to match the repository after running `npm run e2e`. This is the result of (a) including time stamps and (b) writing back the verification result in `verify`.

Finally I'm moving an `rmSync()` call to keep the JSON TLA+ for debugging purposes, in case Apalache fails.


solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -23,17 +23,24 @@ export async function extractContractCall(
...
     // https://developers.stellar.org/network/horizon/api-reference/resources/operations/object/invoke-host-function
 
-    // extract the call data: transaction hash, parameters
+    // Extract the call data: time, transaction hash, parameters
+
+    // In Soroban, `env.ledger().timestamp()` "[r]eturns a unix timestamp for when the ledger was closed":
+    // https://docs.rs/soroban-sdk/latest/soroban_sdk/ledger/struct.Ledger.html#method.timestamp
+    // according to [this](https://stellar.stackexchange.com/questions/1852/transaction-created-at-and-ledger-close-time),
+    // the transaction object's `created_at` equals the ledger's `closed_at`.
+    const timestamp = new Date(op.created_at).getTime() / 1000
```
Kukovec, line 43: Do we need the `/1000` here? I understand that it says [here](https://docs.rs/soroban-sdk/latest/soroban_sdk/ledger/struct.Ledger.html#method.timestamp) that the time is given in seconds, but in a degenerate case it could happen that `Date(t1).getTime()` and `Date(t2).getTime()` are different in `ms` units but same in `s` units due to truncation.

Kukovec (approved):
LGTM

solarkraft/src/fetcher/callDecoder.ts
thpani: Horizon only gives us seconds precision, it's just that `Date` (which we use for parsing) internally represents milliseconds.

We could leave it as milliseconds here, but I think it's preferable to have parity between the spec and the contract code (in case there are any constants).

I agree though that this looks confusing. I'll add a comment for clarification.

Issue #94, opened 2024-06-05
The absolutely necessary things to finish the activation phase


konnov:
We have a lot of things implemented. However, we still need a few things to deliver a demo of the MVP:

 - [x] the TLA+ specification of the timelock monitor running with `solarkraft fetch + verify`: #99
 - [x] whenever a violation is found, `solarkraft verify` should invoke the alert contract: #95 
 - [x] cut a short screencast demonstrating `solarkraft fetch; solarkraft verify` and that the alerts are published in the Stellar block explorer
 - [x] announce the demo at the Stellar discord
 - [x] finish the blogpost
 - [x] announce it on twitter

PR #96, opened 2024-06-05
Automatic alert trigger

### LINKED ISSUE #95: Automatically invoke the alert contract upon verification failure.

Kukovec:
closes #95 

Adds `--alert` flag to `verify`. If supplied with a valid contract id, will submit the verification result to the alert contract deployed with the given ID _on the TESTNET_. If not supplied, `verify` behaves as it did before.

Also removes some misc. stdout noise in `invokeAlert.ts`


solarkraft/src/invokeAlert.ts
```diff
@@ -69,23 +69,19 @@ export async function invokeAlert(
 
     // Let's see the base64-encoded XDR of the transaction we just built.
```
thpani, line 70: I think the comment here is oudated

solarkraft/src/main.ts
```diff
@@ -76,6 +76,11 @@ const verifyCmd = {
                 desc: 'Monitors to check against',
                 type: 'array',
                 demandOption: true,
+            })
+            .option('alert', {
+                desc: 'Alert contract ID. WIll not alert if this paremeter is not provided.',
```
thpani, line 81:
```suggestion
                desc: 'Alert contract ID. Will not alert if this parameter is not provided.',
```

solarkraft/src/verify.ts
```diff
@@ -103,6 +108,50 @@ function apalacheCheck(monitor: string): Result<ApalacheResult> {
     }
 }
 
+// For now, our solution is TESTNET-ONLY!, so
```
thpani, line 111: Maybe file an issue with the enhancement label?
That's how I'm keeping track of stuff that's fine for now, but should eventually be resolved.

solarkraft/src/verify.ts
```diff
@@ -103,6 +108,50 @@ function apalacheCheck(monitor: string): Result<ApalacheResult> {
...
+}
+
+// Trigger the alert contract, if an alert contract ID is provided, otherwise do nothing
+async function conditionalAlert(
+    monitorAnalysisStatus: MonitorAnalysisStatus,
+    txHash: string,
+    alertContractID?: string
+): Promise<void> {
+    if (alertContractID !== undefined) {
+        if (!/^[A-Z0-9]{56}$/g.test(alertContractID)) {
```
thpani, line 131: I believe all contract IDs start with a `C`.

solarkraft/src/verify.ts
```diff
@@ -185,6 +234,14 @@ export function verify(args: any) {
         'unverified' // only returned if array is empty
     )
 
+    const monitorStatus =
+        verificationStatus === 'ok'
+            ? MonitorAnalysisStatus.NoViolation
+            : MonitorAnalysisStatus.Violation
```
thpani, line 240: Hmmm, so this would mark it as violation even if we have some internal error.

Looking at the code block above, I think it should reduce to `unverified` if there's an internal error on one of the monitors and no other monitor fails. Then we don't need any mapping from strings, but can reduce to a `MonitorAnalysisStatus` above directly.

solarkraft/src/verify.ts
```diff
@@ -185,6 +234,14 @@ export function verify(args: any) {
...
+
+    // PR comment: call only if Violation?
+    conditionalAlert(monitorStatus, args.txHash, args.alert)
```
thpani, line 243:
```suggestion
    conditionalAlert(monitorStatus, args.txHash, args.alert)
```

I mean, you say in the alert contract that it gets called for each analysis: 🤷 

https://github.com/freespek/solarkraft/blob/417ec2e5ca5fe99cf4255bc047e6b24abb7a7bda/ContractExamples/contracts/alert/src/lib.rs#L26

thpani (approved):
LGTM, some residual comments.

Looking at it again, I'm not sure the way we reduce analysis results for multiple monitors is correct.

solarkraft/src/verify.ts
Kukovec: I do, and it does the way it's implemented now, but if you look at the text Igor wrote in #94 it only mentions violations.

thpani: Igor wrote #94 in 2 minutes, those are not meant as precise descriptions.

Just be consistent across the verifier and the contract.

solarkraft/src/verify.ts
thpani: > you'd still want the alert contract to go off

But it will go off if you call it on each analysis, and if it changes from `NoViolation` to `Unknown` isn't that precisely what happens?

thpani: > or migrate all 'ok' | 'fail' | 'unknown' to MonitorAnalysisStatus everywhere, which would also be an option

that's what I meant, but I haven't checked how big the impact would be

Kukovec: Not too drastic, and it's probably cleaner to have one single way of enumerating outcomes, than 2 and convert between. I'll try to include this in this PR, unless I notice it getting extremely out-of-scope.

solarkraft/src/verify.ts
konnov: We should ask the actual potential users. I can see how only reporting failures would make sense, but also how reporting both would make sense too.

Kukovec: > We should ask the actual potential users. I can see how only reporting failures would make sense, but also how reporting both would make sense too.

The current implementation always calls the contract, which then emits different kinds of events, depending on the status. A user may choose to subscribe to any subset of those. The only time the alert contract stores data is when violations occur.

Kukovec:
A small heads up RE the diff of the alst commit: I've unified `MonitorAnalysisStatus` and `VerificationStatus` everywhere into a `VerificationStatus` enum, so it's just a cascade of name changes

PR #98, opened 2024-06-06
Update devcontainer to soroban-cli 21-rc1

thpani:
Update devcontainer to soroban-cli 21-rc1


konnov (approved):
Thanks for fixing it!

Issue #99, opened 2024-06-10
Verifiable timelock monitor


Kukovec:
Create a monitor that is compatible with our instrumentation, based on #37

PR #100, opened 2024-06-10
Instrumentable spec

Kukovec:
Implements half of #99 (only `deposit` for now)

Makes the following changes:
 - Adds the two contracts (`timelock`, `token`) from `soroban_examples` under `ContractExamples/contracts`
 - Adds a monitor based on #37 under `docs/timelock`, together with a type hints file to use with `fetch`
 - Adds two scripts which:
    - initialize `Docker` and populate `soroban` with aliases 
    - Automate the deployment of the two contracts and basic invocations, ending in a `deposit` call
 - Changes instrumentation, s.t.:
    - `env` now gets instrumented with two additional fields, tracking which values were set in storage before and after the transaction    
    - The order of conjuncts in `Next` is changed, so that it initializes before calling the method operator, avoiding use-before-assignment in the case where the method operator references primed variables
    - The way `Variants!Variant` arguments are passed is fixed.


konnov:
we have to add proper attribution to the files we copy from elsewhere

ContractExamples/scripts/init-timelock.sh
```diff
@@ -0,0 +1,32 @@
...
+    --testnet \
+    --enable-soroban-rpc;
+
+soroban network add \
+    --rpc-url $RPC \
+    --network-passphrase "Test SDF Network ; September 2015"\
+    testnet
+
+soroban keys generate alice --network testnet;
+soroban keys generate bob --network testnet;
```
thpani, line 28: This is project-level configuration that's shared with all the other contracts here.

I don't think we should re(set) it inside a `timelock`-specific script.

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+soroban keys address $BOB || (echo "add the account $BOB via soroban keys generate"; exit 1)
+
+set -x
+
+TOKEN="`
+    soroban contract deploy \
+    --source-account alice \
+    --network testnet\
+    --rpc-url $RPC\
+    --network-passphrase "Test SDF Network ; September 2015"\
```
thpani, line 27: There's no need to override these, you're already giving them in `soroban network add`

```suggestion
```

Kukovec:
Just a heads up regarding the commit you're about to see (TLA noise). Apalache translates `UNCHANGED x` not into `x' = x`, but into `x' := x` (using the assignment operator). This becomes a problem, because `x' = x /\ UNCHANGED x` trigger an "assigneed twice" exception in Apalache, since left-to-right analysis necessitates that the first conjunct, `x' = x`, be taken as an assignment as well. Semantically, there is no change, but since I had to change the conjunt order in `Next` instrumentation, I have to update the current test monitors using UNCHANGED as well.

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+    --wasm $SCRIPT_DIR/../contracts/token/target/wasm32-unknown-unknown/release/soroban_token_contract.wasm\
+    `";
```
thpani, line 29: I don't see what the other double quotes give us here?
In any case, I wouldn't nest double quotes.

```suggestion
TOKEN=$(
    soroban contract deploy \
    --source-account alice \
    --network testnet\
    --rpc-url $RPC\
    --network-passphrase "Test SDF Network ; September 2015"\
    --wasm $SCRIPT_DIR/../contracts/token/target/wasm32-unknown-unknown/release/soroban_token_contract.wasm\
    )
```

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+
+set -x
+
+TOKEN="`
+    soroban contract deploy \
+    --source-account alice \
+    --network testnet\
+    --rpc-url $RPC\
+    --network-passphrase "Test SDF Network ; September 2015"\
+    --wasm $SCRIPT_DIR/../contracts/token/target/wasm32-unknown-unknown/release/soroban_token_contract.wasm\
```
thpani, line 28:
```suggestion
    --network testnet \
    --rpc-url $RPC \
    --network-passphrase "Test SDF Network ; September 2015" \
    --wasm $SCRIPT_DIR/../contracts/token/target/wasm32-unknown-unknown/release/soroban_token_contract.wasm
```

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "Timelock contract ID: $TIMELOCK"
+
+soroban contract invoke \
+    --id $TIMELOCK\
+    --source alice \
+    --network testnet \
+    -- \
+    deposit \
+    --from alice \
+    --token $TOKEN\
```
thpani, line 67:
```suggestion
    --id $TIMELOCK \
    --source alice \
    --network testnet \
    -- \
    deposit \
    --from alice \
    --token $TOKEN \
```

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+    --amount 1 \
+    --claimants "[\"$(soroban keys address bob)\"]"\
+    --time_bound '{"kind": "After", "timestamp": 0}'
+
+
+
+
+
+
+
```
thpani, line 77: Can we remove the 7 lines of whitespace here?

```suggestion
    --claimants "[\"$(soroban keys address bob)\"]"\
    --time_bound '{"kind": "After", "timestamp": 0}'
```

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -0,0 +1,77 @@
...
+    --id $TIMELOCK\
+    --source alice \
+    --network testnet \
+    -- \
+    deposit \
+    --from alice \
+    --token $TOKEN\
+    --amount 1 \
+    --claimants "[\"$(soroban keys address bob)\"]"\
+    --time_bound '{"kind": "After", "timestamp": 0}'
```
thpani, line 70: Maybe use a more realistic whitespace here?

thpani (commented):
First set of comments – I haven't looked at the TLA+ yet

1. I don't understand why you diverge from the existing scripts in `ContractExamples/scripts/` so much. They would be much easier to use if they all follow the same function + style.

2. There's a [LICENSE](https://github.com/stellar/soroban-examples/blob/f595fb5df06058ec0b9b829e9e4d0fe0513e0aa8/LICENSE) on `solarkraft-examples` that we have to comply with.

ContractExamples/scripts/init-timelock.sh
```diff
@@ -0,0 +1,32 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+RPC="https://soroban-testnet.stellar.org:443"
+
+docker run --rm -d \
+    -p 8000:8000 \
+    --name stellar \
+    stellar/quickstart:testing \
+    --testnet \
+    --enable-soroban-rpc;
```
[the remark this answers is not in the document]
Kukovec: Fair point, this just stuck around when things were being copied

thpani:
> This becomes a problem, because `x' = x /\ UNCHANGED x` triggers an "assigneed twice" exception in Apalache, since left-to-right analysis necessitates that the first conjunct, `x' = x`, be taken as an assignment as well. Semantically, there is no change, but since I had to change the conjunct order in `Next` instrumentation

Looks to me like you should emit `UNCHANGED x`, not `x' = x` in the first place.

Otherwise,

> I have to update the current test monitors using UNCHANGED as well.

it sounds like we'd have to discourage use of `UNCHANGED` in the monitor specs?
Not particularly fond of that idea.

Kukovec:
>...
> it sounds like we'd have to discourage use of `UNCHANGED` in the monitor specs? Not particularly fond of that idea.

I'll try do dig up the conversation in the Apalache repo explaining why we translate unchanged as an assignment, but from memory, the TLDR is that `x' = x` is a regular condition, whereas `UNCHANGED x` is an actionable declaration of the shape of the successor state, and since Apalache has assignments, which base TLA lacks, the `UNCHANGED -> :=` translation felt like it made sense at the time, even though it's working against us here.

ContractExamples/scripts/prepare-timelock.sh
thpani: Sorry, I meant more realistic timestamp 😅

doc/timelock/deposit.tla
```diff
@@ -0,0 +1,102 @@
...
+
+\* @type: ($depositArgs) => Bool;
+MustHold_deposit_BalanceRecordCorrect(args) ==
+    /\ Balance'.token = args.token
+    /\ Balance'.amount = args.amount
+    /\ Balance'.time_bound = args.time_bound
+    /\ Balance'.claimants = args.claimants
+
+\* Everything below is deterministic, and will be generated automatically
+\* For now, we encode this manually
```
thpani, line 48: Can we make this more visible?

The point is that monitors are small, so we should clearly separate the manual instrumentation part.

thpani (commented):
second set of comments

konnov (commented):
I had a very quick view at the PR:

1. Why don't we link the soroban examples via github modules? Otherwise, it would be real hard to keep code in sync, when soroban-examples are updated. Also, what's the point of copying the code from another project? We would have to take care of the licenses and other uninteresting stuff.
2. I don't get the discussion about `UNCHANGED`. Yes, we thought that forcing `UNCHANGED` to be an assignment was a good idea, but why do we introduce two assignments in the first place?

ContractExamples/scripts/prepare-timelock.sh
Kukovec: "After 0" should mean that the claimants can claim whenever, which I thought would be easy to follow-up test. If you hardcode any fixed timestamp you either a) functionally get the same thing as 0 if that time has already passed, or b) have to actually wait before calling `claim`.

doc/timelock/state.tla
```diff
@@ -0,0 +1,68 @@
...
+instance_has(key, env) ==
+    key \in env.storage_before
+
+\* @type: ($storageKey, $env) => Bool;
+next_instance_has(key, env) ==
+    key \in env.storage_after
+
+\* @type: ($storageKey, $env) => Bool;
+instance_set(key, env) ==
+    env.storage_after \ env.storage_before = {key}
```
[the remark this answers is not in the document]
Kukovec: So the original code here was
```
instance_storage' = instance_storage \union {key}
```
My understanding of the intent here was that the interpretation of this was "exactly `key` gets set" (since "key is set" could have been tested just via `key \in instance_storage'`), and I rewrote it that way, because `A = B \cup {x}` is true iff `A` differs from `B` by exactly `x` _or_ if they both contain `x`, which I assumed to have been an oversight.
Your suggestion, `x \in A\B` would evaluate to true if `A \ B` equaled `{x, y, z, ...}` too (which may or may not have been the actual intent). Perhaps @andrey-kuprianov can comment best on what the intended semantics here were.

doc/timelock/deposit.tla
```diff
@@ -0,0 +1,109 @@
+(*
+ * Deposit method monitors for the Timelock contract
+ * 
+ * Per our original discussion in https://github.com/freespek/solarkraft/blob/32f48e5f26e4cce76df4703a1b29fd1752cd1677/doc/case-studies/timelock/README.md
+ * the MustHold_<method>_<id>, MustPass_<method>_<id>, and MustFail_<method>_<id> ( MustRevert here ) invariants could, in the future,
+ * get automatically composed into Must<X>_<method>, and ultimately into a single Inv_<method> operator, since the
+ * compositions are fully deterministic.
+ *
+ * As this monitor is written to illustrate our intended approach, these compositions are written manually from their respective 
+ * sub-components, follwoing the shape utlined in the above README.
```
thpani, line 10: Please get a spell checker


```suggestion
 * sub-components, following the shape outlined in the above README.
```

konnov:
I don't want to block this PR on `UNCHANGED`. IMO, let's open an issue on `UNCHANGED`, maybe we even have to revisit Apalache on that. If we can avoid issues with `UNCHANGED` in the current examples, let's do that.

.github/workflows/main.yml
```diff
@@ -57,6 +57,9 @@ jobs:
           - stable
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - run: git submodule init && git submodule update --remote --recursive
```
thpani, line 62: If you look at the last build run, `submodules: true` already initializes and fetches the submodule:

https://github.com/freespek/solarkraft/actions/runs/9465023037/job/26073622926?pr=100#step:2:422

```suggestion
```

thpani (commented):
I guess this is fine – not the kind of PR I would've liked to review multiple times.

The discussion on `UNCHANGED` is lost on me, maybe @konnov still wants to comment.

thpani (approved):
Unblocking this given Igors comment

PR #105, opened 2024-06-12
Correctness fix in default status

Kukovec:
Found this while doing the demo, since `VerificationStatus::Unknown` absorbs `NoViolation`, the initial fold-value must be a `NoViolation`, since otherwise `Unknown` (initial) + `NoViolation` (actual result) give an `Unknown`, instead of a `NoViolation`


thpani (approved):

thpani:
I think you'll have to rebase on main to get the tests working

PR #116, opened 2024-08-27
Clean up dead code

thpani:
Housekeeping: remove unused dead code


konnov (approved):
It's a good idea to clean up this code before starting the new phase!

PR #117, opened 2024-08-27
Move non-commands to subdir

thpani:
Small refactoring that clears up the `./src/` root, to only contain entry points to the main commands `fetch`, `list`, and `verify`.

All other source files are moved to subdirectories.


konnov (approved):
LGTM

PR #127, opened 2024-09-23
Update contract IDs and transaction hashes after testnet reset

thpani:
After the recent testnet reset, redeploy the test contracts and update contract IDs and transaction hashes.


konnov (approved):
LGTM

PR #129, opened 2024-09-26
Stable integration tests for the transaction fetcher

### LINKED ISSUE #56: Develop more stable integration tests for the transaction fetcher
As noted in #55, the integration test may stop working when the contract ledger history disappears, due to a TTL, or due to a testnet reset. We need an integration test that populates the network with data and then evaluates it. One blocker towards that is that [Stellar binaries](https://github.com/stellar/binaries/), used in Github actions, are not supporting v21 yet.

thpani:
Our fetcher integration test so far relies on hard-coded transaction hashes that were pre-populated by the `setter-popluate.sh` script. Those tx hashes need to be manually updated whenever the testnet resets, or when the setter contract is updated.

This PR introduces programmatic population of the deployed setter contract to achieve a more stable integration test: On each integration test run, a fresh copy of the setter contract is created from its on-chain WASM code (there is a 1:n relationship between WASM code and contracts IDs). Each test then invokes a setter function of the contract, followed by fetching the respective transaction and verifying the call decoder output as before.

This could be further optimized for performance (i.e., only repopulate the contract when necessary), but for now a stable integration test is more important.

Closes #56

### DIFF
--- a/solarkraft/test/integration/callDecoder.test.ts
+++ b/solarkraft/test/integration/callDecoder.test.ts
@@ -1,26 +1,47 @@
 /**
  * Integration tests that ensure that our contract decoder works correctly on the testnet.
- * These tests use hardcoded ledger entries that are created with:
  *
- * `ContractExamples/scripts/setter-populate.sh`
+ * First deploys a fresh copy of the setter contract (`ContractExamples/contracts/setter`)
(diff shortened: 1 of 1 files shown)


konnov (approved):
This is cool! I did not know that one could deploy a contract from WASM saved on-chain :)

PR #130, opened 2024-09-26
Support `ledgerEntryRemoved`

### LINKED ISSUE #122: Handle `ledgerEntryRemoved`
I suspect that this would be triggered by unsetting temporary or permament storage:

https://github.com/freespek/solarkraft/blob/204be6f39d70a6cc48b0c3f87322dcde8944596f/solarkraft/src/fetcher/callDecoder.ts#L148-L150

Triage and fix.

thpani:
So far, we did not support the `ledgerEntryRemoved` ledger change type in the fetcher.
It occurs when a persistent or temporary entry is removed via `env.storage()._().remove(key)`.

Turns out it already works out-of-the-box, but this PR adds a test (including a removal function in the setter contract).

Closes #122

### DIFF
--- a/ContractExamples/contracts/setter/src/lib.rs
+++ b/ContractExamples/contracts/setter/src/lib.rs
@@ -112,6 +112,14 @@ impl SetterContract {
         env.storage().instance().get(&MY_BOOL).unwrap_or(false)
     }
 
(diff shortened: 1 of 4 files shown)


konnov (approved):
Really cool!

PR #131, opened 2024-09-27
Add support for fetching failed transactions

### LINKED ISSUE #61: [Epic] M2: Handle failed transactions
Currently reverted transactions are ignored by the fetcher+checker.

We should retrieve and check them to achieve full support for `MustRevert_`.

thpani:
Closes #61

Extract failed transaction (in addition to successful ones) in the fetcher.

The implementation itself is straight-forward – testing is a bit more complicated, because we need to reliably produce 2 transactions that fail concurrently and are not discarded by simulation upfront. We achieve this by submitting 2 tx from different accounts.

ℹ️ Probably easier to review commmit-by-commit.


solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -50,6 +50,7 @@ export async function extractContractCall(
     // `created_at` and `env.ledger().timestamp()` have seconds-precision. We divide by 1000 here since `Date` internally
     // represents the timestamp as milliseconds.
     const timestamp = new Date(op.created_at).getTime() / 1000
+    const transaction_successful = op.transaction_successful
```
konnov, line 54: This is a long name. How about `tx_ok` (or `txOk` since we have `txHash`)?

solarkraft/test/integration/callDecoder.test.ts
```diff
@@ -40,8 +40,9 @@ const WASM_HASH =
 // contract ID of the deployed setter contract (will be set up by `before()`)
 let CONTRACT_ID: string
 
-// a random keypair to sign transactions
-const sourceKeypair = Keypair.random()
+// random keypairs to sign transactions
```
konnov, line 43: how about:

```
// random keypairs to sign transactions (we need two accounts to produce concurrent transactions)
```

konnov (approved):
Very nicely done! I thought it would require more code.

solarkraft/src/fetcher/callDecoder.ts
thpani: I've shortened it to `txSuccess` (I don't think `txOk` conveys enough meaning)

solarkraft/test/integration/callDecoder.test.ts
thpani: good point!

PR #132, opened 2024-10-08
Fix storage loading

thpani:
Fix a type annotation and loading of storage `MultiContractStorage` maps:

`Immutable.fromJS` deeply converts all JS objects to Maps.
For deserializing storage, we want the target object to instead be
```
Map -> JS object {instance: Map, persistent: Map, temporary: Map}
```
Thus, we provide a custom loader instead.


konnov (approved):
Great fix!

Issue #133, opened 2024-10-10
Previous audits as case studies


thpani:
We can use previous audit reports as case studies; here are some audit reports that have been produced as part of the [Soroban Audit Bank](https://stellar.org/blog/developers/the-soroban-audit-bank-fostering-a-secure-smart-contract-ecosystem).

| Category     | Project | Auditors |
|--------------|---------|----------|
| **Lending**  | [Blend Protocol](https://github.com/blend-capital/blend-contracts) | [OtterSec](https://ottersec.notion.site/Sampled-Public-Audit-Reports-a296e98838aa4fdb8f3b192663400772), [Certora](https://certora.cdn.prismic.io/certora/65b2a04e615e73009ec3ef2d_BlendCertoraReport.pdf) |
| – _flash loan_ | [Xycloans](https://github.com/xycloo/xycloans) | [OtterSec](https://ottersec.notion.site/Sampled-Public-Audit-Reports-a296e98838aa4fdb8f3b192663400772) |
| **DEX / AMM**| [CometDEX](https://github.com/CometDEX/comet-contracts-v1) | [OtterSec](https://ottersec.notion.site/Sampled-Public-Audit-Reports-a296e98838aa4fdb8f3b192663400772) |
|              | Aquarius | [CoinFabrik](https://github.com/CoinFabrik/coinfabrik-audit-reports/blob/main/Aquarius/CoinFabrik-Aquarius_AMM_Audit-March_2024.pdf) |
|              | Cables Finance | [Certora](https://certora.cdn.prismic.io/certora/ZhbgfzjCgu4jzvBP_Certora_Report_Cables_V2.pdf), [Halborn](https://github.com/HalbornSecurity/PublicReports/blob/master/Soroban%20Smart%20Contract%20Audits/Shift_Markets_Cables_DEX_Soroban%20Smart_Contract_Security_Assessment_Report_Halborn_Final.pdf) |
|              | [Allbridge Estrela](https://github.com/allbridge-io/dex-soroban-contracts) | [Quarkslab](https://allbridge.io/assets/docs/reports/24-03-1573-REP-Allbridge-Estrela-Review-V1.1.pdf) |
[...]

PR #135, opened 2024-10-14
Introduce TypeScript monitors

### LINKED ISSUE #134: Introduce Solarkraft monitors in TypeScript
In order to evaluate how we write monitors, we need to iterate quickly based on several [case studies](https://github.com/freespek/solarkraft/issues/133).

The current TLA+ instrumentation is a bit too brittle to iterate quickly, and we have not fully understood how to translate Soroban's heterogeneous/duck-typed storage into TLA+ types.

Thus, we introduce a monitor driver for writing monitors in TypeScript.

thpani:
In order to evaluate how we write monitors, we need to iterate quickly based on several [case studies](https://github.com/freespek/solarkraft/issues/133).

The current TLA+ instrumentation is a bit too brittle to iterate quickly, and we have not fully understood how to translate Soroban's heterogeneous/duck-typed storage into TLA+ types.

Thus, we introduce a monitor driver for writing monitors in TypeScript.

Using the TS monitor driver introduced in this PR, writing a monitor in TypeScript is very simple: Create a TS class that extends `SolarkraftJsMonitor` and provide methods corresponding to the contract functions. In the body of those functions, use `reverts_if` and `succeeds_with` methods to specify the conditions under which the contract should revert or succeed. The monitor is then (programmatically) verified by instantiating the monitor class and calling the `verify` method with the hashes of the transactions to be verified.
For an example, see f6e994e7e605015575663d82e536b2969333e977.

Closes #134


thpani:
@konnov This is a rather large change-set, so I recommend reviewing by commit.

solarkraft/src/verify_js.ts
```diff
@@ -0,0 +1,329 @@
...
+
+/**
+ * Return a condition that checks if the given token amount has been transferred from one account to another.
+ */
+export function tokenTransferred(
+    env: Env,
+    token: string,
+    from: string,
+    to: string,
+    amount: number
```
konnov, line 169: would not it be better to use big integers right from the start? Otherwise, we would not be able to test amount that do not fit into `number`.

solarkraft/test/e2e/verify_js_example.ts
```diff
@@ -0,0 +1,120 @@
...
+    token: string
+    amount: number
+    claimants: string[]
+    time_bound: TimeBound
+}
+
+/* the monitor */
+
+class TimelockMonitor extends SolarkraftJsMonitor {
+    deposit(
```
konnov, line 51: I like this. One idea: When you write `deposit`, this may be confusing, as you do not describe the behavior of `deposit`. Would it make sense to have a name like `on_deposit`, `before_deposit`, or `deposit_ensures`?

solarkraft/src/types.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license
+ * [Apache-2.0](https://github.com/freespek/solarkraft/blob/main/LICENSE)
+ */
+/**
+ * Types shared across the project
+ */
+import { Either } from '@sweet-monads/either'
+
+export type Result<T> = Either<string, T>
```
konnov: I like that a lot!

konnov (approved):
I like this idea a lot. Obviously, we need more examples and iterations to figure out, what would be needed to specify monitors right in JS. However, it is a very good approach to avoiding the complexity of Rust and TLA+.

solarkraft/src/verify_js.ts
thpani: True, I also encountered this while working on an additional case study.

I have to do a bit of digging to find out where we introduce `number`. `js-stellar-sdk` actually should give us `bigint`s, so they are probably lost somewhere in serialization.

thpani: Since I already have another PR wedged inbetween, I'm fixing this in a separate one: #140

solarkraft/test/e2e/verify_js_example.ts
thpani: Good point! I think `on_deposit` would work well.

I've put it in #141 to follow up on.

PR #140, opened 2024-10-18
Use bigint everywhere

thpani:
To avoid mixing `number` and `bigint`, we should use bigint for all numerical contract data.

We previously lost them by writing JSON to disk, this rectifies it by instructing `JSONbigint` to always load numerical types as native bigint.


konnov (approved):
Very nice refactoring!

PR #144, opened 2024-10-21
Add the Xycloans case study

### LINKED ISSUE #142: Add case study #1
Identify a case study from #133 and implement a monitor

thpani:
Add a further case study, based on TypeScript monitors.

It reproduces a known and fixed criticial vulnerability exposed by a rounding error identified by OtterSec as finding `OS-XYC-ADV-00` in [this audit report](https://github.com/xycloo/xycloans/blob/main/audits/osec.pdf).

Closes #142


solarkraft/test/e2e/verify_js_example_xycloans.ts
```diff
@@ -0,0 +1,134 @@
...
+            )
+        })
+    }
+
+    deposit(from: string, amount: bigint) {
+        this.succeeds_with((env: Env) => {
+            const token = env.storage().instance().get('TokenId')
+
+            // side-effects: we track deposits locally to compute rewards later
+            this.shares = this.shares.update(from, 0n, (x) => x + amount)
```
konnov, line 66: This looks like a very fragile solution to me. What if we have two `succeeds_with` conditions that both update `this.shares`?

konnov (approved):
Amazing example! We should get some feedback from the Xycloans team on that

solarkraft/test/e2e/tla/.stor/CBBDVFSEI26N7DJBPMVQDNWX2GD7RF2PC2HYMUVKGU72NCFAKERFYIKL/477912/entry-49ce70a0fe73f9b815f50dbe13403608242cec0770ac32eb24d1a425e39467be.json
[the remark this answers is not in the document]
thpani: It's used in the e2e test, which just does `verify`, not `fetch`.
We could of course also `fetch` there, but the e2e are already quite slow.

solarkraft/test/e2e/verify_js_example_xycloans.ts
thpani: We could move it outside the `succeeds_with`. WDYT?

konnov: yes, I believe it's how it should be done conceptually. Kind of like one update per tx.

thpani: Uh, but here's the thing: we're passing `Env` to `succeeds_with` not to the containing method. However, the second side-effect below accesses storage via `env`.

We're doing this because `reverts_if` and `succeeds_with` receive different `Env`s: `reverts_if` only has access to the pre-state, while `succeeds_with` receives pre- and post-states.

thpani: I will think about it a bit more, but maybe we need to leave this as a future design consideration.

ContractExamples/scripts/xycloans-populate.sh
```diff
@@ -0,0 +1,71 @@
...
+
+cd ${dir}/../xycloans
+
+NET=testnet
+(soroban network ls | grep -q $NET) || (echo "add testnet via soroban network"; exit 1)
+
+ADMIN=admin
+soroban keys address $ADMIN || (echo "add the account $ADMIN via soroban keys generate"; exit 1)
+
+ACCOUNT=alice
```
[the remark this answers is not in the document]
thpani: Done in 6b5870a

PR #147, opened 2024-11-25
upgrade dependencies to avoid a high

konnov:
`npm` reported a high vulnerability in our dependencies. Upgrading them.

PR #148, opened 2024-11-25
add an npm audit workflow

konnov:
Adding a workflow to automatically audit npm dependencies


konnov:
Once #147 is fixed, this should work

thpani (approved):

PR #149, opened 2024-11-27
minimal TLA+ spec of the xycloans monitor

konnov:
This is basically the TLA+ specification after `verify_js_example_xycloans.ts`. Apalache produces examples for `MCxycloans_monitor.tla`. However, we probably need more constraints in the spec, as the environment gets a complete update every time. We have to figure out how to produce meaningful examples.

### DIFF
--- a/doc/case-studies/xycloans/MCxycloans_monitor.tla
+++ b/doc/case-studies/xycloans/MCxycloans_monitor.tla
@@ -83,8 +83,12 @@ Next ==
                 \/ update_fee_rewards([ env |-> env, call |-> UpdateFeeRewards(addr), status |-> success ])
             \* Propagate the new storage. For value generation, we go over all addresses, not subsets.
             /\ storage' = IF success THEN new_stor ELSE storage
-                /\ success => tx.env.old_storage = storage
+
+\* use this falsy invariant to generate examples of successful transactions
+NoSuccessInv ==
+    ~IsCreate(last_tx.call) => ~last_tx.status
 
 \* use this view to generate better test coverage
+\* apalache-mc check --max-error=10 --length=10 --inv=NoSuccessInv --view=View MCxycloans_monitor.tla
 View == <<last_tx.status, last_tx.call>>
 =========================================================================================
--- a/doc/case-studies/xycloans/MCxycloans_monitor.tla
+++ b/doc/case-studies/xycloans/MCxycloans_monitor.tla
@@ -1,0 +1,96 @@
+------------------------------- MODULE MCxycloans_monitor -------------------------------
+(*
+ * An instance of xycloans monitor for model checking and input generation.
+ *
+ * Igor Konnov, 2024
+ *)
+EXTENDS Integers, xycloans_types
+
+\* the set of all possible token amounts
+AMOUNTS == Nat
+\* the contract address for the XY Loans contract
+XYCLOANS == "xycloans"
+\* the token address
+XLM_TOKEN_SAC_TESTNET == "xlm-sac"
+
+\* the pool of addresses to draw the values from
+ADDR == { "alice", "bob", "eve", XLM_TOKEN_SAC_TESTNET, XYCLOANS }
+
+VARIABLES
+    \* @type: $tx;
+    last_tx,
+    \* @type: Str -> Int;
+    shares,
+    \* @type: Int;
+    total_shares,
+    \* @type: Int;
+    fee_per_share_universal
+
+INSTANCE xycloans_monitor
+
+Init ==
+    /\ last_tx = [
+            call |-> Create("any"),
+            status |-> TRUE,
+            env |-> [
+                current_contract_address |-> "any",
+                storage |-> [
+                    self_instance |-> [
+                        FeePerShareUniversal |-> 0,
+                        TokenId |-> ""
+                    ],
+                    self_persistent |-> [
+                        Balance |-> [ addr \in {} |-> 0 ],
+                        MaturedFeesParticular |-> [ addr \in {} |-> 0 ],
+                        FeePerShareParticular |-> [ addr \in {} |-> 0 ]
+                    ],
+                    token_persistent |-> [ Balance |-> [ addr \in {} |-> 0 ] ]
+                ],
+                old_storage |-> [
+                    self_instance |-> [
+                        FeePerShareUniversal |-> 0,
+                        TokenId |-> ""
+                    ],
+                    self_persistent |-> [
+                        Balance |-> [ addr \in {} |-> 0 ],
+                        MaturedFeesParticular |-> [ addr \in {} |-> 0 ],
+                        FeePerShareParticular |-> [ addr \in {} |-> 0 ]
+                    ],
+                    token_persistent |-> [ Balance |-> [ addr \in {} |-> 0 ] ]
+                ]
+            ]
+        ]
+    /\ shares = [ addr \in {} |-> 0 ]
+    /\ total_shares = 0
+    /\ fee_per_share_universal = 0
+
+Next ==
+    \* generate some values for the storage
+    \E fpsu \in AMOUNTS, tid \in { "", XLM_TOKEN_SAC_TESTNET }:
+      \E b, mfp, fpsp, tb \in [ ADDR -> AMOUNTS ]:
+        LET storage == [
+                self_instance |-> [ FeePerShareUniversal |-> fpsu, TokenId |-> tid ],
+                self_persistent |-> [ Balance |-> b, MaturedFeesParticular |-> mfp, FeePerShareParticular |-> fpsp ],
+                token_persistent |-> [ Balance |-> tb ]
+            ]
+            env == [
+                current_contract_address |-> XYCLOANS,
+                storage |-> storage,
+                old_storage |-> last_tx.env.storage
+            ]
+        IN
+        \E method \in { "initialize", "deposit", "borrow", "update_fee_rewards" }:
+            \E addr \in ADDR, amount \in AMOUNTS, success \in BOOLEAN:
+                LET call ==
+                    CASE method = "initialize" -> Initialize(XLM_TOKEN_SAC_TESTNET)
+                      [] method = "deposit" -> Deposit(addr, amount)
+                      [] method = "borrow" -> Borrow(addr, amount)
+                      [] method = "update_fee_rewards" -> UpdateFeeRewards(addr)
+                IN
+                LET tx == [ env |-> env, call |-> call, status |-> success ] IN
+                \/ initialize(tx)
+                \/ deposit(tx)
+                \/ borrow(tx)
+                \/ update_fee_rewards(tx)
+
+=========================================================================================
--- a/doc/case-studies/xycloans/MCxycloans_monitor.tla
+++ b/doc/case-studies/xycloans/MCxycloans_monitor.tla
@@ -39,18 +39,18 @@ Init ==
             TokenId |-> ""
         ],
         self_persistent |-> [
-            Balance |-> [ addr \in {} |-> 0 ],
-            MaturedFeesParticular |-> [ addr \in {} |-> 0 ],
-            FeePerShareParticular |-> [ addr \in {} |-> 0 ]
+            Balance |-> [ addr \in ADDR |-> 0 ],
+            MaturedFeesParticular |-> [ addr \in ADDR |-> 0 ],
+            FeePerShareParticular |-> [ addr \in ADDR |-> 0 ]
         ],
         token_persistent |-> [ Balance |-> [ addr \in {} |-> 0 ] ]
     ]
     IN
     /\ last_tx = [
-            call |-> Create("any"),
+            call |-> Create(XYCLOANS),
             status |-> TRUE,
             env |-> [
-                current_contract_address |-> "any",
+                current_contract_address |-> XYCLOANS,
                 storage |-> init_stor,
                 old_storage |-> init_stor
             ]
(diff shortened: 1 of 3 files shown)


doc/case-studies/xycloans/xycloans_monitor.tla
```diff
@@ -0,0 +1,138 @@
...
+    /\ IsBorrow(tx.call)
+    \* `FeePerShareUniversal` has been updated correctly
+    /\ succeeds_with(tx,
+           expected_fee_per_share_universal = tx.env.storage.self_instance.FeePerShareUniversal)
+    \* the receiver paid the expected fee to the pool
+    /\ succeeds_with(tx, old_token_balance(tx.env, call.receiver_id) >= call.amount)
+    /\ LET a == call.receiver_id IN
+       succeeds_with(tx, token_balance(tx.env, a) = old_token_balance(tx.env, a) - call.amount)
+    /\ LET a == tx.env.current_contract_address IN
+       succeeds_with(tx, token_balance(tx.env, a) = old_token_balance(tx.env, a) + call.amount)
```
thpani, line 111: I don't like that I have to substitute `a` in my brain. We should find a better name.

Also, repeating it is confusing, as it refers to different addresses.

doc/case-studies/xycloans/xycloans_monitor.tla
```diff
@@ -0,0 +1,138 @@
...
+        expected_fee == div_ceil(call.amount * STROOP, 12500000000)
+        expected_fee_per_share_universal ==
+            tx.env.old_storage.self_instance.FeePerShareUniversal
+                + div_floor(expected_fee * STROOP, total_shares)
+    IN
+    /\ IsBorrow(tx.call)
+    \* `FeePerShareUniversal` has been updated correctly
+    /\ succeeds_with(tx,
+           expected_fee_per_share_universal = tx.env.storage.self_instance.FeePerShareUniversal)
+    \* the receiver paid the expected fee to the pool
```
thpani, line 106: This comment refers to all 3 conjuncts below, right?

It's a bit hard to match, maybe we should indent the respective `succeeds_with` conjuncts if they represent a single concept.

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,97 @@
...
+                    CASE method = "Initialize" -> Initialize(XLM_TOKEN_SAC_TESTNET)
+                      [] method = "Deposit" -> Deposit(addr, amount)
+                      [] method = "Borrow" -> Borrow(addr, amount)
+                      [] method = "UpdateFeeRewards" -> UpdateFeeRewards(addr)
+                IN
+                LET tx == [ env |-> env, call |-> call, status |-> success ] IN
+                /\  \/ initialize(tx)
+                    \/ deposit(tx)
+                    \/ borrow(tx)
+                    \/ update_fee_rewards(tx)
```
thpani, line 91: Instead of a tagged union for the different calls, we could lift the `$call` parameters directly into the operator arguments, and apply them here?

doc/case-studies/xycloans/xycloans_monitor.tla
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * Igor Konnov, 2024
+ *)
+
+EXTENDS Integers, xycloans_types
+
+STROOP == 10000000
+
+CONSTANT
+    \* The token address for the XY Loans contract.
```
thpani, line 20: xyc

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,97 @@
...
+(*
+ * An instance of xycloans monitor for model checking and input generation.
+ *
+ * Igor Konnov, 2024
+ *)
+EXTENDS Integers, xycloans_types
+
+\* the set of all possible token amounts
+AMOUNTS == Nat
+\* the contract address for the XY Loans contract
```
thpani, line 11: xyc

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,97 @@
...
+                      [] method = "UpdateFeeRewards" -> UpdateFeeRewards(addr)
+                IN
+                LET tx == [ env |-> env, call |-> call, status |-> success ] IN
+                /\  \/ initialize(tx)
+                    \/ deposit(tx)
+                    \/ borrow(tx)
+                    \/ update_fee_rewards(tx)
+                \* propagate the new storage
+                \* TODO: new_stor may contain updates to a subset of addresses
+                /\ storage' = IF success THEN new_stor ELSE storage
```
thpani, line 94: Is the TODO already taken care of by assigning the whole `storage`?

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,97 @@
...
+                /\ success => tx.env.old_storage = storage
```
thpani, line 95: Isn't the RHS always the case, by construction?

thpani (commented):
Looks good!

I like this a lot, it looks rather clean for a TLA+ spec.

Although it feels like it would be a lot of work to generate all the boilerplate programmatically and/or to train engineers to write the right code.

I've left some comments, mostly at the spec level. I have to think about the state machine a bit more.

doc/case-studies/xycloans/MCxycloans_monitor.tla
konnov: oops, copilot thought it was XY  :))

doc/case-studies/xycloans/MCxycloans_monitor.tla
konnov: Yeah, I am still undecided about what to do here. Perhaps, passing the whole storage solves the issue of partiality, at least, for value generation.

doc/case-studies/xycloans/xycloans_monitor.tla
```diff
@@ -0,0 +1,138 @@
...
+
+\* @type: ($tx, Bool) => Bool;
+reverts_with(tx, cond) == ~tx.status => cond
+\* @type: ($tx, Bool) => Bool;
+succeeds_with(tx, cond) == tx.status => cond
+\* @type: (Str -> a, Str, a) => a;
+get_or_else(map, key, default) ==
+    IF key \in DOMAIN map THEN map[key] ELSE default
+\* integer division with rounding up
+div_ceil(a, b) == a + (b - 1) \div b
```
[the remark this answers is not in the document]
konnov: oh, I think you are right :D

doc/case-studies/xycloans/MCxycloans_monitor.tla
konnov: good point, I have simplified it significantly

doc/case-studies/xycloans/MCxycloans_monitor.tla
konnov: true, this is a leftover from earlier modeling

doc/case-studies/xycloans/MCxycloans_monitor.tla
konnov: I've added a comment instead of TODO

konnov:
> Although it feels like it would be a lot of work to generate all the boilerplate programmatically and/or to train engineers to write the right code.

Yeah, some parts are actually generated by copilot from the typescript version pretty well. The issue is, of course, that we have to show copilot a few examples of how to do it right.

konnov:
> I've left some comments, mostly at the spec level. I have to think about the state machine a bit more.

Addressed all of the comments

konnov:
@thpani this one is fixed

doc/case-studies/xycloans/xycloans_types.tla
```diff
@@ -0,0 +1,87 @@
...
+        | UpdateFeeRewards({ addr: Str})
+    ;
+
+    Finally, a transaction is:
+
+    @typeAlias: tx = {
+        env: $env,
+        call: $call,
+        status: Bool
+    };
```
thpani, line 49: Looking at it a second time, it feels a bit weird to have such a deeply-nested record in TLA+ only to emulate the Soroban API.

In the end, you're only using `call` and `status` in the state machine, which could be lifted into TLA+ variables? `env.storage` is already tracked in a state variable, which leaves only `env.current_contract_address`.

doc/case-studies/xycloans/xycloans_monitor.tla
```diff
@@ -0,0 +1,152 @@
...
+    /\ IsInitialize(tx.call)
+    /\ reverts_if(tx, tx.env.old_storage.self_instance.TokenId /= "")
+    /\ succeeds_with(tx, tx.env.storage.self_instance.TokenId = XLM_TOKEN_SAC_TESTNET)
+    \* these conditions are not required by a monitor, but needed to avoid spurious generated values
+    /\ succeeds_with(tx,
+        tx.env.storage.self_instance.FeePerShareUniversal = tx.env.old_storage.self_instance.FeePerShareUniversal)
+    \* these conditions are not required by a monitor, but needed to avoid spurious generated values
+    /\ succeeds_with(tx,
+        tx.env.storage.self_persistent = tx.env.old_storage.self_persistent)
+    /\ last_tx' = tx
```
thpani, line 69: Should we move the `last_tx` assignment into the `MC_` file?

I think it should happen at the same level as the `storage` assignment.

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,98 @@
...
+            env |-> [
+                current_contract_address |-> XYCLOANS,
+                storage |-> init_stor,
+                old_storage |-> init_stor
+            ]
+        ]
+    /\ shares = [ addr \in {} |-> 0 ]
+    /\ total_shares = 0
+    /\ fee_per_share_universal = 0
+    /\ storage = init_stor
```
thpani, line 61: Can we order the variables and group them into
- those that would be present in any monitor (`last_tx`, `storage`), and
- those that specifically achieve stateful tracking for Xycloans (`shares`, `total_shares`, `fee_per_share_universal`).

A comment to explain why the latter ones are there would also help.

thpani (approved):
Looks good, I had some more ideas reading it the second time.

Let's see how this works out with generating values!

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,98 @@
...
+                    [ Balance |-> b, MaturedFeesParticular |-> mfp, FeePerShareParticular |-> fpsp ],
+                token_persistent |-> [ Balance |-> tb ]
+            ]
+            env == [
+                current_contract_address |-> XYCLOANS,
+                storage |-> new_stor,
+                old_storage |-> storage
+            ]
+        IN
+        \E addr \in ADDR, amount \in AMOUNTS, success \in BOOLEAN:
```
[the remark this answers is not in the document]
thpani: I meant the lines 79++

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,98 @@
...
+            /\  \/ initialize([ env |-> env, call |-> Initialize(XLM_TOKEN_SAC_TESTNET), status |-> success ])
+                \/ deposit([ env |-> env, call |-> Deposit(addr, amount), status |-> success ])
+                \/ borrow([ env |-> env, call |-> Borrow(addr, amount), status |-> success ])
+                \/ update_fee_rewards([ env |-> env, call |-> UpdateFeeRewards(addr), status |-> success ])
+            \* Propagate the new storage. For value generation, we go over all addresses, not subsets.
```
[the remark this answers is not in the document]
konnov: Removed

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,98 @@
...
+    \* @type: Str -> Int;
+    shares,
+    \* @type: Int;
+    total_shares,
+    \* @type: Int;
+    fee_per_share_universal,
+    \* Keep track of the current storage,
+    \* which can be only changed by a successful transaction.
+    \* @type: $storage;
+    storage
```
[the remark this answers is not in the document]
konnov: Yeah, my idea was that `last_tx` contains only partial storage updates like in Soroban, and `storage` contains the full storage. However, this did not work well so far.

doc/case-studies/xycloans/xycloans_types.tla
konnov: In general, `env` may contain other things such as the timestamp. So I would prefer to have `env` to be clearly separated from the other things. Do you like to introduce `env`, `call`, and `status` to be three parameters instead of `tx`? They are all essential to a monitor.

doc/case-studies/xycloans/xycloans_monitor.tla
konnov: Yep, totally

doc/case-studies/xycloans/MCxycloans_monitor.tla
```diff
@@ -0,0 +1,98 @@
...
+        self_persistent |-> [
+            Balance |-> [ addr \in ADDR |-> 0 ],
+            MaturedFeesParticular |-> [ addr \in ADDR |-> 0 ],
+            FeePerShareParticular |-> [ addr \in ADDR |-> 0 ]
+        ],
+        token_persistent |-> [ Balance |-> [ addr \in ADDR |-> 0 ] ]
+    ]
+    IN
+    /\ last_tx = [
+            call |-> Create(XYCLOANS),
```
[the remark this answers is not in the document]
konnov: Good point. I've renamed it to `Constructor`, which makes more sense in the light of the recent changes to Soroban

konnov:
I've fixed a few things after your comments. Still undecided about how to properly deal with partial storage. Let's see whether we discover it on our way

doc/case-studies/xycloans/MCxycloans_monitor.tla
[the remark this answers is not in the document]
thpani: Aaah, makes sense. Let's see where this goes

doc/case-studies/xycloans/xycloans_types.tla
thpani: Let's keep it as is and iterate from there

PR #150, opened 2024-11-29
fixing the CI + workflow for deploying contracts to testnet

konnov:
I did not expect that it will take me so much time (like a week) to fix the CI with all the recent Soroban changes. What we have here:
 - [x] An update to stellar/rs-soroban-sdk 22.0.0-rc.3.1 in the examples
 - [x] An update to `@stellar/stellar-sdk` v13.0
 - [x] A downgrade to `rustc` 1.81.0, otherwise, a few examples break on deployment with references-something-we-love-rust, see stellar/rs-soroban-env#1469. 
 - [x] The deployment scripts fixed and working with v22.0.0 on the testnet
 - [x] Calling `__constructor` instead of `initialize` in timelock, since [cap-0058](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0058.md)
 - [x] Since the Horizon API does not have `result_meta_xdr` any longer (two weeks earlier than announced), using `getTransaction` from the `rpc.Server`. Now we use both Horizon API and RPC API. Perhaps, we could switch to RPC completely in the future.
 - [x] Add workflow `testnet.yml` to deploy all the contracts we use, including the old version of `xycloans`. This workflow is only triggered on-demand or once per week, so we don't have to spam the testnet on every commit.

Optimally, it would be great to have a smaller PR, but all these things pulled one another.

### DIFF
--- a/ContractExamples/.cargo/config.toml
+++ b/ContractExamples/.cargo/config.toml
@@ -1,0 +1,5 @@
+[build]
+rustflags = ["--cfg", "target_family=\"wasm\""]
+
+[target.wasm32-unknown-unknown]
+rustflags = ["-C", "target-feature=+bulk-memory"]
--- a/ContractExamples/.cargo/config.toml
+++ b/ContractExamples/.cargo/config.toml
@@ -1,5 +1,2 @@
-[build]
-rustflags = ["--cfg", "target_family=\"wasm\""]
-
 [target.wasm32-unknown-unknown]
 rustflags = ["-C", "target-feature=+bulk-memory"]
--- a/ContractExamples/.cargo/config.toml
+++ b/ContractExamples/.cargo/config.toml
@@ -1,2 +1,0 @@
-[target.wasm32-unknown-unknown]
-rustflags = ["-C", "target-feature=+bulk-memory"]
--- a/ContractExamples/Cargo.toml
+++ b/ContractExamples/Cargo.toml
@@ -7,7 +7,7 @@ members = [
 ]
 
 [workspace.package]
-version = "22.0.0-rc.3"
+version = "22.0.0-rc.3.1"
 edition = "2021"
 
 [workspace.dependencies]
--- a/ContractExamples/Cargo.toml
+++ b/ContractExamples/Cargo.toml
@@ -12,7 +12,6 @@ edition = "2021"
 
 [workspace.dependencies]
 soroban-sdk = { version = "22.0.0-rc.3.1" }
-soroban-token-sdk = { version = "22.0.0-rc.3.1" }
 
 [profile.release]
 opt-level = "z"
--- a/ContractExamples/Cargo.toml
+++ b/ContractExamples/Cargo.toml
@@ -7,7 +7,7 @@ members = [
 ]
 
 [workspace.dependencies]
-soroban-sdk = "21.6.0"
+soroban-sdk = "22.0.0-rc.3"
 
 [profile.release]
 opt-level = "z"
--- a/ContractExamples/Cargo.toml
+++ b/ContractExamples/Cargo.toml
@@ -27,9 +27,3 @@ lto = true
 [profile.release-with-logs]
 inherits = "release"
 debug-assertions = true
-
-[profile.release.package.soroban-env-host]
-debug = false
-
-[profile.release.package.stellar-xdr]
-debug = false
--- a/ContractExamples/Cargo.toml
+++ b/ContractExamples/Cargo.toml
@@ -6,6 +6,10 @@ members = [
   "soroban-examples/timelock/"
 ]
 
+[workspace.package]
+version = "22.0.0-rc.3"
+edition = "2021"
+
 [workspace.dependencies]
 soroban-sdk = "22.0.0-rc.3"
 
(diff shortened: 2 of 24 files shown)


ContractExamples/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[target.wasm32-unknown-unknown]
+rustflags = ["-C", "target-feature=+bulk-memory"]
```
thpani, line 2: Why do we need this?

ContractExamples/Cargo.toml
```diff
@@ -23,3 +28,9 @@ lto = true
 [profile.release-with-logs]
 inherits = "release"
 debug-assertions = true
+
+[profile.release.package.soroban-env-host]
```
thpani, line 32: I don't see us using these additional profiles anywhere. 

Are they somehow expected?

ContractExamples/Cargo.toml
```diff
@@ -6,8 +6,13 @@ members = [
...
 ]
 
+[workspace.package]
+version = "22.0.0-rc.3.1"
+edition = "2021"
+
 [workspace.dependencies]
-soroban-sdk = "21.6.0"
+soroban-sdk = { version = "22.0.0-rc.3.1" }
+soroban-token-sdk = { version = "22.0.0-rc.3.1" }
```
thpani, line 15: Why do we need this additional dep?

ContractExamples/scripts/prepare-timelock.sh
```diff
@@ -9,36 +9,35 @@
...
 
-set -x
+#set -x
 
-TOKEN=$(
-    soroban contract deploy \
+./scripts/build.sh
+
+stellar contract deploy \
+    --salt `date +%s` \
```
thpani, line 25: Do we need to provide the salt?

ContractExamples/scripts/xycloans-populate.sh
```diff
@@ -12,60 +12,66 @@ set -e
 
 dir=$(cd `dirname $0`; pwd -P)
 
+# see: https://github.com/stellar/rs-soroban-sdk/pull/1353
+rustc --version | grep 1.81.0 || (echo "Run: rustup default 1.81.0"; exit 1)
```
thpani, line 16: Can't we use build.sh here?

ContractExamples/scripts/xycloans-populate.sh
```diff
@@ -12,60 +12,66 @@ set -e
...
+
 cd ${dir}/../xycloans
+patch --forward -p1 <../patches/xycloans.diff || (echo "patch already applied?")
+patch --forward -p1 <../patches/xycloans2.diff || (echo "patch already applied?")
+
+stellar contract build
```
thpani, line 22: build.sh?

solarkraft/package.json
```diff
@@ -41,8 +41,8 @@
     "test": "mocha --loader=ts-node/esm --exclude test/e2e/**/*.test.ts --exclude test/integration/**/*.test.ts test/**/*.test.ts"
   },
   "dependencies": {
-    "@stellar/stellar-base": "^11.0.1",
-    "@stellar/stellar-sdk": "^12.2.0",
+    "@stellar/stellar-base": "^12.1.1",
```
thpani, line 44: We have to nuke stellar-base at some point. 
The docs explicitly say you shouldn't use both -sdk and -base

solarkraft/src/fetcher/callDecoder.ts
```diff
@@ -80,9 +81,16 @@ export async function extractContractCall(
     // https://developers.stellar.org/network/horizon/api-reference/resources/transactions/object
     const height = tx.ledger_attr
 
+    // Since Horizon does not return tx.result_meta_xdr anymore, make another query with getTransaction
+    const txDetails = await rpcServer.getTransaction(txHash)
+    if (txDetails.status == rpc.Api.GetTransactionStatus.NOT_FOUND) {
```
thpani, line 86: Does this every happen? 
Feels like we should output a warning or something if we have to skip a tx that's expected to exist.

thpani (commented):
Looks good, thank you for taking care of this! 🙏🏻 

I just have a few small questions about small fragments that aren't clear:

ContractExamples/.cargo/config.toml
konnov: this came from Stella. I think we can remove it

ContractExamples/Cargo.toml
konnov: For some reason, the token example is using it as a dependency. Let me remove it

ContractExamples/scripts/prepare-timelock.sh
konnov: Nope. That was a desperate attempt to remove deployment bugs

solarkraft/src/fetcher/callDecoder.ts
konnov: This should not happen. However, a missing transaction does not have a field. Let's issue a warning

ContractExamples/Cargo.toml
konnov: I believe it was also Stella

thpani (approved):

solarkraft/package.json
konnov: Good point. It is actually not needed

PR #154, opened 2024-12-12
Repairing E2E tests

konnov:
Since the testnet is getting periodically reset, this PR:
 
 - [x] fixes the current integration and e2e tests
 - [x] stores hard-coded addresses in the population scripts, so we don't have to do anything by hand ever again

@thpani I don't know why so many entries in `test/tla/.stor` have changed. Shall we delete some of them?


.github/workflows/testnet.yml
```diff
@@ -6,11 +6,13 @@ on:
     # This cron expression triggers the workflow at 8:00 UTC every day
     - cron: "8 0 * * *"
   # uncomment to debug in a PR
-  #pull_request:
+  pull_request:
```
thpani, line 9: Intentional that you uncommented this?

.github/workflows/testnet.yml
```diff
@@ -22,6 +24,26 @@ jobs:
...
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "temurin" # See 'Supported distributions' for available options
+          java-version: "19"
+
+      - name: Install apalache for e2e
+        run: |
+          gh release download --repo apalache-mc/apalache --pattern apalache.tgz
+          tar -xvzf apalache.tgz --directory . > /dev/null
+          echo "`pwd`/apalache/bin" >> $GITHUB_PATH
```
thpani, line 37: Where is GITHUB_PATH used? Maybe add a comment?

.github/workflows/testnet.yml
```diff
@@ -22,6 +24,26 @@ jobs:
...
+
+      - name: Install solarkraft for e2e
+        run: |
+          cd solarkraft
+          npm install --dev
```
thpani, line 42: Since this is e2e, shall we drop `--dev`?

ContractExamples/scripts/alert-deploy.sh
```diff
@@ -25,4 +29,7 @@ set -x
 ./scripts/build.sh
 
 stellar contract deploy --wasm target/wasm32-unknown-unknown/release/alert.wasm \
-      --source $ACCOUNT --network $NET
\ No newline at end of file
+      --source $ACCOUNT --network $NET | tee >.alert.id
+
+CONTRACT_HASH=`head -n 1 .alert.id`
```
thpani, line 34: I think this is the (contract) address, not a (WASM) hash?

ContractExamples/scripts/setter-populate.sh
```diff
@@ -27,43 +31,52 @@ set -x
...
 
+HEIGHT=`curl -s https://horizon-testnet.stellar.org/ | jq .history_latest_ledger`
+echo "export const SETTER_HEIGHT = $HEIGHT" >> $HARDCODED
+
 stellar contract deploy --wasm target/wasm32-unknown-unknown/release/setter.wasm \
-      --source $ACCOUNT --network $NET | tee >.setter.id
+      --source $ACCOUNT --network $NET 2>.setter.err | tee >.setter.id
+grep "wasm hash" err | sed 's/.*hash \([0-9a-f]*\)/\1/' > .setter.hash
+
+CONTRACT_HASH=`head -n 1 .setter.id`
```
thpani, line 41: I think we should call this `CONTRACT_ADDR`

solarkraft/test/e2e/generated/alertHardcoded.ts
```diff
@@ -0,0 +1,3 @@
+// AUTO-GENERATED by alert-deploy.sh
+export const ALERT_CONTRACT_HASH =
```
thpani, line 2: Do we need to commit the hard-coded files?
afaiu, CI would generate them?

solarkraft/test/e2e/generated/setterHardcoded.ts
```diff
@@ -0,0 +1,6 @@
+// AUTO-GENERATED by setter-populate.sh
+export const SETTER_HEIGHT = 53424
```
thpani, line 2: Sim, do we need to commit this?

solarkraft/test/e2e/verify.test.ts
```diff
@@ -85,31 +93,62 @@ describe('verify', () => {
...
-    txs.forEach((tx) => {
-        it(`verifies the setter contract (tx ${tx})`, function (done) {
-            this.timeout(150_000)
-            spawn(
-                `solarkraft verify --home test/e2e/tla/ --txHash ${tx} --monitor test/e2e/tla/setter_mon.tla`
-            )
-                .wait('[OK]')
-                .run(done)
-        })
+    it('fetches and verifies', function (done) {
```
thpani, line 96:
```suggestion
    it('fetches and verifies the setter contract', function (done) {
```

solarkraft/test/e2e/verify.test.ts
```diff
@@ -85,31 +93,62 @@ describe('verify', () => {
...
+                        console.log(`checked ${e.txHash} of ${ncalls}`)
+                        if (nchecked >= ncalls) {
+                            // technically, we may have a race condition
+                            // between spawning and finishing verification
+                            assert(ncalls > 0, 'no transactions checked')
+                            // we have to tell mocha that we are done
+                            done()
+                        }
+                    })
+                ncalls++
```
thpani, line 134: This looks super fishy. You're incrementing `ncalls` outside the callback.
Won't you call `done()` early if the callback executes before the outer loop makes another call?
It looks like you're assuming that the outer loop executes faster than the callback, but it may be completely arbitrary.

I think the proper way to do this is to return a promise and wait for `Promise.all()`.

solarkraft/test/e2e/verify.test.ts
thpani, line 96: Since you now fetch before verifying, can we remove some of the fetched entries in `test/tla/.stor`?

thpani (commented):
Generally looks good, I just made a few comments.

My only big concern is what you're doing with `nchecked`/`ncalls` in `solarkraft/test/e2e/verify.test.ts`. It looks like an anti-pattern.

.github/workflows/testnet.yml
konnov: Yes, I will commit it again before merging. Just wanted to make sure that this workflow is going through

.github/workflows/testnet.yml
konnov: done

.github/workflows/testnet.yml
konnov: we are still using mocha in the setup, so it needs `--dev`

ContractExamples/scripts/alert-deploy.sh
konnov: yes, this is confusing. The tutorials seem to call it a contract id. Let's rename to `CONTRACT_ADDR`.

solarkraft/test/e2e/generated/setterHardcoded.ts
konnov: well, I imagine that we do not push new addresses often, just when the testnet is reset, though the deployment is tried every night. We will debug the tests more often.

solarkraft/test/e2e/verify.test.ts
konnov: yes, I was fighting with this a lot. You are right. Ideally, we would like to wait for all the promises to end. However, `nexpect` is not using promises. Instead, it uses some kind of `node` tasks. I have no idea how to wait there. So it's the best I could come up with. It may terminate earlier.

solarkraft/test/e2e/verify.test.ts
konnov: Good point. I've managed to remove most of them except a few that are still used

thpani (approved):

solarkraft/test/e2e/verify.test.ts
konnov: I have found a way to do it properly, similar to how you did it by dynamically introducing tests in a loop. The same trick works with a `before` hook!

thpani (approved):

